PR #71, opened 2024-04-13
Cleanup pjt dev

irgolic:
- migrate to poetry
- move one of the scripts into a test
- add CI for tests, format, lint, and typecheck
- reformat the code with ruff, fix lint errors
- set up ignore-revs to skip the linting/formatting commits in git blame
- initialize a makefile with [test, type, format, lint, all] targets
- general cleanup (remove .idea, rename readme)


xinaesthete:
Just one `# type: ignore` left in the code... if you have any bright ideas @irgolic that'd be interesting, I'm not sure if the type system is expressive enough to deal with this, perhaps it'd be clearer for the reader with a slightly more verbose version anyway, but I kinda like the terse version:

```python
def compute_vcf_end(df: pd.DataFrame) -> pd.DataFrame:
    """
    Compute the end position of the variant determined from 'POS', 'REF' and 'ALT'.

    This is added as a column 'END' in the given DataFrame.
    """

    def varlen(s: str) -> int:
        return max([len(s) for s in s.split(",")])

    # python/panda types aren't clever enough to easily deal with this (tbf fairly gnarly syntax)
    # df[['REF', 'ALT']] resolves to `Series[Unknown]`
    df["END"] = df["POS"] + df[["REF", "ALT"]].applymap(varlen).max(axis=1)  # type: ignore
    return df
```

xinaesthete:
(we should possibly be using some kind of library for VCF files anyway / approaching the whole thing differently, but that's another story)

irgolic:
> Just one `# type: ignore` left in the code... if you have any bright ideas @irgolic that'd be interesting, I'm not sure if the type system is expressive enough to deal with this, perhaps it'd be clearer for the reader with a slightly more verbose version anyway, but I kinda like the terse version:
> 
> ```python
> def compute_vcf_end(df: pd.DataFrame) -> pd.DataFrame:
>     """
>     Compute the end position of the variant determined from 'POS', 'REF' and 'ALT'.
> 
>     This is added as a column 'END' in the given DataFrame.
>     """
> 
>     def varlen(s: str) -> int:
>         return max([len(s) for s in s.split(",")])
> 
>     # python/panda types aren't clever enough to easily deal with this (tbf fairly gnarly syntax)
>     # df[['REF', 'ALT']] resolves to `Series[Unknown]`
>     df["END"] = df["POS"] + df[["REF", "ALT"]].applymap(varlen).max(axis=1)  # type: ignore
>     return df
> ```

If I remove the `# type: ignore` I'm not getting any errors @xinaesthete

xinaesthete:
I don't get any errors when I run pyright e.g. via `make all`, but the `pylance` extension in VSCode (using Pylance language server) is... One way or another, it'd be good to have the configuration of language server agree with other scripts / CI.

irgolic:
> I don't get any errors when I run pyright e.g. via `make all`, but the `pylance` extension in VSCode (using Pylance language server) is... One way or another, it'd be good to have the configuration of language server agree with other scripts / CI.

Pylance uses pyright under the hood (they're both microsoft). What pyright version is your pylance running?

xinaesthete:
I'm not sure how to determine that. Also possible that a more complete restart of vscode might help.

xinaesthete:
I see this in Python Language Server output

```
2024-04-16 11:27:03.728 [info] (Client) Pylance async client (2024.4.1) started with python extension (2024.4.1)
2024-04-16 11:27:04.493 [info] [Info  - 11:27:04] (2785) Pylance language server 2024.4.1 (pyright version 1.1.356, commit 6652c4a8) starting
```

irgolic:
Sorry, can't reproduce it; I went through these possibilities https://github.com/microsoft/pyright/issues/4784#issuecomment-1471896936

There will always be some discrepancies between tooling and IDE. Just pick a source of truth (`e.g., make type`) and abide by it/run it in CI, even if other tools throw errors.

xinaesthete:
ok, thanks for looking into it anyway.

xinaesthete:
@irgolic  The real issue was that `applymap` in Pandas is deprecated, and replacing it with `map` appears to keep everyone happy. The type of `Series[Unknown]` is very fair given that the only thing known about `df` is there's all kinds of potential for unchecked exceptions if that function is called out of context, in which case I think the raised errors will be equally as useful as any explicit checks we'd pollute the function itself with.

xinaesthete:
Also I think I still have some issues with the Sphinx documentation generation which I must remember to fix - also needed to review that anyway to make it parse all my TypeScript code as the existing config uses `jsdoc`.

PR #80, opened 2024-07-18
Jh dev

jayeshhireox:


xinaesthete:
I'm not sure why `version: '3.8'` was added back to `docker-compose*.yml`; we get a warning that it's deprecated and from what I've read it should just be removed.

PR #88, opened 2024-08-22
TIFF Upload UI with Metadata Visualization

aleql:


aleql:
The following changes have been made to the branch:

First, I’ve removed the logic for adding or updating data sources from the frontend and adjusted the workflow accordingly with the backend implementation proposed. The current flow for uploading TIFF images is as follows:

- The first request is sent to the `/upload` endpoint, with the payload containing the TIFF image file and the project name.

- Upon a successful file upload, a second request is triggered to /add_or_update_image_datasource. Since this endpoint is still being defined, the payload currently includes the TIFF image metadata and the datasource name selected by the user. 
- 
- Once the request to this endpoint is completed successfully, a call to `state_saved` in `ChartManager` is executed.

To prevent inconsistencies, I’ve linked the second request as a callback to the success of the first, ensuring that the datasource is only updated after a successful file upload. Separating these processes also allows us to focus on improving the robustness of the file upload itself.

Additionally, I’ve made adjustments to `ProjectContext` to provide both the project name and root for the upload request in `FileUploadComponent`. I also fixed the TIFF image size issue when the image is expanded

xinaesthete (approved):

PR #95, opened 2024-10-14
viv experimental

xinaesthete:
Switching to experimental fork of vivjs so we can start to use new deck.gl / luma.gl & associated community-layers etc.


xinaesthete:
I don't think there's anything missing in terms of functionality or outstanding bugs - notwithstanding `VivVolume` which I haven't tested and isn't used outside `obvios_index.ts` / expect to be super-ceded by a more react-y version.

PR #96, opened 2024-10-15
Enhance accessibility, readability, and color palette consistency in CSS

aleql:


xinaesthete:
I'm not so sure about some of these changes, can you do a bit more testing in dark mode? Some of the new borders around certain elements seem a bit distracting to me (actually even in light mode I'm not so sure about the border around `.ciview-menubar`).

From my point of view, I think I'd be more interested in seeing dark-mode implemented in the catalog view - and at some point we'll want to review how to straighten out vanilla-CSS / material theme / tailwind etc.

src/utilities/css/ContextMenu.css
```diff
@@ -141,12 +141,14 @@
...
-  background-image: var(--menu-gradient);
+.ciview-menu-bar {
+  flex: 0 0 auto;
+  padding: 5px 25px;
+  background-color: var(--menu_bar_color);
   color: var(--text_color);
-
+  margin-top: 5px; /* Add a top margin */
+  border: 2px solid var(--border_menu_bar_color);
+  box-shadow: 2px 4px 8px rgba(0, 0, 0, 0.2);
```
xinaesthete, line 151: I think this looks good on `border-bottom`, other borders seem to be what was distracting to me particularly in dark-mode (but also subtly in light mode)
```suggestion
  border-bottom: 2px solid var(--border_menu_bar_color);
```

src/charts/css/charts.css
```diff
@@ -175,14 +190,15 @@ select, input, textarea {
 .ciview-chart-titlebar{
     display:flex;
     padding:4px;
-    background: var(--title_bar_color);
     color: var(--text_color);
+    background-color: var(--background_color);
     flex:0 0 auto;
     font-size:14px;
+    border: 1px solid #dfdfdf; /* Add a subtle border */
 }
```
xinaesthete, line 198:
```suggestion
    border: 1px solid rgba(120,120,120, 0.2); /* Add a subtle border */
```
this looks subtle to me on both light & dark

xinaesthete:
I'm going to merge then apply the tweaks I suggested.

## Before:
Screenshot 2024-10-19 at 11 38 24
Screenshot 2024-10-19 at 11 38 11

## After:
Screenshot 2024-10-19 at 11 36 35
Screenshot 2024-10-19 at 11 36 43

nb - the scatterplot graphics (not changed with this PR) are pretty illegible with defaults in light-mode; should be addressed soon hopefully.

Also note that changing the menu-bar stuff to be rendered with React / Material is on the agenda so we can start having a nicer view select drop-down etc.

src/utilities/css/ContextMenu.css
xinaesthete: not sure how that `box-shadow` removal got in there, not intended.

PR #97, opened 2024-10-15
Tiff upload UI enhanced error handling

aleql:


xinaesthete:
Should all be fine - I might want to try to do a bit more testing of re-throwing the error in `createLoader` - I have a feeling I may have put something like that in at some point and may have had a reason to go back to just returning the null data, but I think it should be ok.

PR #102, opened 2024-10-24
UI Catalog improved error handling

aleql:
I’ve included these changes to enhance error handling, along with improvements to the Catalog. Now, the Catalog displays a new modal that provides detailed information whenever an error occurs, offering clearer feedback for failed requests. 

Also, I’ve addressed these cases by updating the file upload management in the FileUploadComponent. Specifically, I’ve added error handling for 403 Forbidden errors, ensuring the system properly handles situations where a file is read-only.


src/catalog/hooks/useProjects.ts
```diff
@@ -133,68 +204,134 @@ const useProjects = () => {
...
-
-      setProjects(prevProjects =>
-        prevProjects.map(project =>
-          project.id === id ? { ...project, name: newName } : project
-        )
-      );
-
+  
+      if (response.status === 404) {
+        throw new Error('Project not found or has been deleted');
```
xinaesthete, line 239: Minor point: maybe we could have a dictionary somewhere outside mapping status code to error message, then this part can just look up from there & throw as appropriate - would just save a little bit of repetition here and structure things a bit more.

xinaesthete:
I haven't actually tested yet, but couldn't see anything that looked too risky to merge - hopefully not tempting fate :)

I'll have a think about sorting out our various branches soon...

Cheers,
Peter

PR #103, opened 2024-10-29
UI catalog project list view

aleql:
The following updates are included in this PR:
- Added a button to the dashboard interface to switch between grid and list views for projects. This button is positioned next to the sorting options dropdown.
- Projects can now be displayed in a list with the following columns: Name, Project Owner, Last Modified Date, Project Type, and an actions button.
- Each project row retains the same functionality as the project cards, offering access to the project and the following modals: Project Info, Rename Project, Delete Project, and Change Project Access.
- Improved project sorting functionality with support for sorting projects by "last modified" and "name" in both ascending and descending order.

### DIFF
--- a/src/catalog/Dashboard.tsx
+++ b/src/catalog/Dashboard.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from "react";
+import React, { useCallback, useMemo, useState } from "react";
 import {
     Add,
     Search,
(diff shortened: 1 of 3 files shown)


xinaesthete:
Looks good, thank you @aleql it's coming together quite nicely.

I think I noticed one or two issues that may have actually already been present before: When handling "create project" response, you check for `response.status === 201` explicitly, meaning that the user gets an error even though the server response was ok - I think we should use `response.ok` there. Also it looks like the formatting is two-spaces in that file (which I somewhat wish I'd set as standard, but will make for some noise at some point). Lastly, I suspect that if you ran the biome linter it might complain about not always using `useCallback` etc - having functions declared directly in components can lead to issues with excessive rendering I think.

Sorry for the niggles - must also sort out our branches hopefully soon...

PR #107, opened 2024-11-07
Added enhanced file upload support to FileUploadDialogComponent

aleql:
- Enhanced and standardized file type recognition within the FileUploadComponent, making it more robust.

- Introduced an interface to simplify future support for additional file types.


xinaesthete:
Am I right in thinking the `maxSize` just makes it so the dropzone won't accept larger files? These files can quite easily be over 10gb so we should either communicate that to the user or review the limit.

PR #110, opened 2024-11-14
Feature/upload 154 

jayeshhireox:
the backend plag has been added to MDVProject constructor , to distinguish the db logic from main api logic without backend.
also, csv upload chart creation part is added.


python/mdvtools/mdvproject.py
```diff
@@ -21,6 +21,7 @@
 import time
 import copy
 from mdvtools.image_view_prototype import create_image_view_prototype
+from mdvtools.charts.table_plot import TablePlot
```
xinaesthete, line 24: I think we probably want to move stuff that uses things like this outside this file

xinaesthete (approved):

PR #115, opened 2024-11-23
column link

xinaesthete:
This is to do with enhancing the way in which columns can be selected - in particular, setting up dynamic links that will respond to operations in linked data-sources etc.


xinaesthete:
@coderabbitai review

src/react/components/ColumnSelectionComponent.tsx
```diff
@@ -1,88 +1,177 @@
...
             .filter((c) => columnMatchesType(c, type))
             ,
         [dataStore, props.exclude, type],
     );
+    // const linkProps = useRowsAsColumnsLinks(); //todo: arbitrary number of links, and subgroups within links
+    if (current_value && typeof current_value !== "string" && !Array.isArray(current_value)) {
+        return <RAComponent {...props} />;
+    };
+    return (
+        <Grid className="w-full items-center" container>
```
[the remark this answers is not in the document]
xinaesthete: The generic constraints are the thing that is without doubt the most confusing thing about this currently, that I'm trying to improve. In this instance, that error in general I don't think will happen - or certainly shouldn't if things are specified and implemented correctly; it's there to help me check at runtime... but any suggestions around refining and better documenting the types would be much appreciated.

nawabfurquan:
@coderabbitai we are trying to come up with a more coherent way of modelling the props for GuiSpec<"column"> | GuiSpec<"multicolumn">and how we do things like setting the state... Can you illustrate how the props passed down through children of <ColumnSelectionSettingGui />are transformed / propagate?

src/react/components/ColumnSelectionComponent.tsx
```diff
@@ -1,89 +1,350 @@
...
+            aria-labelledby={`tab-${index}`}
+            style={{padding: '1em 0.5em', width: "100%"}}
+            {...other}
+            >
+            {value === index && children}
+        </div>
+    )
+};
+
+const LinkTo = observer(<T extends CTypes,>(props: RowsAsColsProps<T>) => {
```
nawabfurquan, line 165: @coderabbitai In this LinkTo component, when I try to assign a new value to props.current_value in a useEffect, I am getting an error saying that current_value is read-only, how to fix it?

xinaesthete: @coderabbitai I agree that we should be using `setSelectedColumn` here - but isn't it true that with `mobx` we might in fact at times mutate props; if you look at any of the components in `SettingsDialogComponent.tsx` you can see examples of this.

xinaesthete: @coderabbitai @nawabfurquan so as I understand it, the reason that `props.current_value` in the `LinkTo` component is readonly is that in `ColumnSelectionSettingGui`, when we do `const props2 = ...`, there is nothing to make that object `observable`... I'm a little hazy, but I think that may have been intentional. Perhaps the logic could be reviewed, and it could certainly be better documented in the code... the logic seems to be that the top-level `ColumnSelectionSettingGui` will wrap the original `observable` props that are given to it, and create a `setSelectedColumn: action` partly in an attempt to avoid situations where we end up mutating some `observable` state in reaction to other observable state changing...

Given that I have comments like

```ts
current_value: props.current_value //! ideally this would also react to changes in the mobx store, why doesn't it?
```

I think what I may need to do is revisit that, perhaps making that particular prop be `@computed`...

xinaesthete: @coderabbitai there is also a bit of a design problem around the way that in `@loadColumnData`, we have an `autorun()` responding to changes in one `observable`, that will go on to call `target.call` which invariably will itself mutate other MobX state... so we get warnings like `"[MobX] Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed. Tried to modify: ObservableObject@21.color_by"`

xinaesthete: @coderabbitai I think what you suggest would've violated the rule about not modifying observable state from a reaction. In this case, it's ok because the second `action` actually is on a callback passed to `ChartManager._getColumnsThen()`, so will be outside the calling scope, but in other situations we need to think of some more general way of modelling this.

src/react/components/ColumnSelectionComponent.tsx
```diff
@@ -1,89 +1,378 @@
...
+    const { linkedDs, link } = props;
+    const { name_column, name, subgroups } = link;
+    const {values, formattedValues} = useLinkTargetValues(props);
+    const { setSelectedColumn, current_value } = props;
+    //@ts-expect-error need to review isMultiType logic
+    const isMultiType = isMultiColumn(props.type);
+
+    // nb, the <"multidropdown" | "dropdown"> type parameter ends up causing problem vs <"multidropdown"> | <"dropdown">
```
xinaesthete, line 220: @coderabbitai can you suggest a better approach for where we have this issue with generics?

src/charts/ChartManager.js
```diff
@@ -179,7 +179,7 @@ export class ChartManager {
          * @typedef {import("@/charts/charts/DataSource")} DataSource
```
xinaesthete, line 179: @coderabbitai this is being resolved as `any` rather than using the `DataSource` type definition from `charts.d.ts` - can we fix that?

xinaesthete (commented):
> @coderabbitai this is being resolved as `any` rather than using the `DataSource` type definition from `charts.d.ts` - can we fix that?

src/charts/ChartManager.js
xinaesthete: @coderabbitai I tried that but it doesn't help. I think it had been working previously, but when we made some changes to fix `typedoc` we had a regression.

nawabfurquan:
@coderabbitai I am facing issue with biome explicitly in this branch in my docker container, it works fine in other branches. It says "Extension 'Biome' is configured as formatter but it cannot format 'TypeScript JSX'-files" even though there is a biome configuration file in it and there is a dependency of biome in the package.json.

nawabfurquan:
@coderabbitai There is a configuration file for biome, it is biome.jsonc. And like I said, biome works fine in other branches, it just doesn't work on this branch.

xinaesthete:
@coderabbitai can you try to review & summarise this... I think I may generate one massive merge commit as it seems a bit hard to chunk the changes & I just want to get it done at this point. If you have a better strategy LMK.

PR #124, opened 2025-01-18
Don't make devcontainer mount ~/mdv multiple times

xinaesthete:
I had added a `volumes` entry in `docker-secrets.yml` (used for configuring local development environment), so that certain scripts and tests using the `~/mdv` convention for project paths would still be able to work in this environment.

I suspect that by having both

```
- ~/mdv:/app/mdv
- ~/mdv:/root/mdv
```

it was leading to files being corrupted.

As such, I have changed the configuration to use `/root/mdv` in the container more consistently. In order for the `mdv_server_app.py` to work, this just required a simple change of `python/mdvtools/config.json` - **however** I'm pretty sure that other places where the container is used will require an equivalent change to their configuration that is not under source control in this repo.

If this is the case, we should review that, as while it should be easy to change the configuration if necessary in order to work with this particular change, it means that it may create friction in general to rollback to different versions / bisect issues etc.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,6 +15,10 @@
         }
     },
     "postStartCommand": "git config --global --add safe.directory /app"
+    //Note: when setting up a new user's development environment, first we had to install
+    // a new WSL distro to clone the code into. Then we had some strange behaviour when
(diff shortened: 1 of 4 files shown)


xinaesthete:
Closing this for now because there are other issues with changing the config for where `~/mdv` is mounted - end up with confused database.

PR #135, opened 2025-02-17
created filter component

xinaesthete:
Work in progress - some known issues with "delete view" functionality as of now.


xinaesthete:
There is a pre-existing niggle with the CSS layout of the menu-bar along the top (`float: right` elements not vertically centred), which is more apparent with this change. This can be addressed separately.

PR #136, opened 2025-02-19
Auth rbac

xinaesthete:


src/login/login.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { Box, Button, Paper, Typography } from "@mui/material";
+import mdvLogo from "./assets/mdv_logo.png";
+
+const Login: React.FC = () => {
+    useEffect(() => {
+        // Check if this is a callback from Auth0
+        const params = new URLSearchParams(window.location.search);
+        if (params.get('callback')) {
+            const base = import.meta.env.DEV ? "http://localhost:5170?dir=/" : "";
+            window.location.href = `${base}/catalog_dev`;
```
xinaesthete, line 12: we must sort out this routing a bit better - not sure we want catalog_dev here if not in vite?

PR #142, opened 2025-03-03
File Upload UI Anndata Nested Processing Update

aleql:
- The frontend is now capable of properly processing these nested AnnData structures, correctly displaying the content on the matrix viewer, including UMAP visualizations.
- Fixed issue present in H5MatrixView component to prevent NaN values from displaying in H5MatrixViewer stats
- Fixed a bug where changing matrix categories could result in no data being displayed if the new category had fewer matrices than the previous active tab index. 

## Summary by CodeRabbit

- **New Features**
  - Enhanced chart display behavior that automatically updates and resets views when new data is detected.
  - Improved numerical formatting for clearer and more consistent presentation of statistical information.
  - Optimized data conversion for accurate handling of diverse value types, leading to more reliable visualizations.

### DIFF
--- a/src/charts/dialogs/H5MatrixViewer.tsx
+++ b/src/charts/dialogs/H5MatrixViewer.tsx
@@ -13,7 +13,7 @@ import {
     Typography,
 } from "@mui/material";
 import type React from "react";
-import { useCallback, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 
 export type Matrix = number[][];
 export type MatrixRecord = Record<string, Matrix | null>;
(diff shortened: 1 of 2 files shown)


xinaesthete:
@coderabbitai review

src/charts/dialogs/H5MatrixViewer.tsx
```diff
@@ -168,6 +184,9 @@ const H5MatrixViewer: React.FC<H5MatrixViewerProps> = ({ matrices }) => {
             const { rows, hasMoreRows, hasMoreCols, colIndices, rowIndices } =
                 visibleMatrixData;
 
+            const formatNumber = (value: number) =>
+                Number.isNaN(value) ? " - " : value.toFixed(4);
```
xinaesthete, line 188: coderabbit is kinda right - I slightly prefer `!Number.isFinite()` here or you may end up with more edge-cases (this hit me in the python code, especially because `Infinity` isn't legal JSON for some unknown reason and the Python `json` module writes it anyway on the basis that "most js engines are ok with it" - "most js engines" doesn't include any prod versions of a modern browser I tested on... I digress).

Having said all that, this is only formatting for display so probably would do less harm anyway.

PR #143, opened 2025-03-07
Rationalise types of state involved in column selection

xinaesthete:
The goal here is to fix bugs and hopefully end up with a more understandable way of representing `ColumnSelectionProps` and related types for `ColumnSelectionSettingGui` / `ColumnSelectionComponent`.


xinaesthete:
@coderabbitai when we call `getParamsGuiSpec(chart)` from `ParamsSettingGui`, it creates an array of settings objects... but the handling of types & states down the line in associated components is somewhat complex and confused. We need to make sure that columns presented to user are filtered to appropriate data-types, and that single vs multiple selection is correctly handled. I hope that we can make the flow and expression of types clearer in the process.

xinaesthete:
@coderabbitai part of the historical reason for the complexity is that I have been trying to marshal ways that older code would express this logic differently in different parts of the application... AddChartDialog with UI for `params` as described in `BaseChart.types` entries, which is where `"_multi_column:number"` etc originate... but there's another issue there with a level of indirection between `"number"` and that `DataTypes` like `"double"` that can be on a column... so I was trying to make some kind of generic for bringing those together - seem to be getting a bit lost along the way. There is also `ColumnSelectionProps.exlude`... but I don't know how important that really is. I just seem to keep going around in circles.

Incidentally, the `@ts-expect-error` comments all indicate places where I am trying to fix things - but quite a few relate to a separate issue with colour related props (which I expect will actually be simpler although it leaks across more of the code).

Anyway, I'm sure there is something simpler I could do to approach this...

Notably, there are basically two 'public' APIs for using column selection - as part of a `SettingsDialogComponent` where props are expressed as `GuiSpec<"column"> | GuiSpec<"multicolumn">` - in which case there's a clear way of discriminating single vs multi, and I think I made the 'multiple' prop in the hope that it would be simpler to use (and there's a simple bug now with that being ignored).

The underlying `ColumSelectionComponent` can be used as a component in arbitrary places, such as the `AddRowComponent` in `SelectionDialogComponent.tsx`, and is also used more generically in for params `AddChartDialogReact.tsx`.

xinaesthete:
@coderabbitai It was probably a mistake because it was contributing to how confusing all of this was, but my earlier implementation was trying to use generics so that the consumers of this code would be able to know, for example, whether the `DataColumn` they get has properties associated with number/categorical columns etc... Ultimately, I would really like the components that are consuming column data to be able to automatically have as much information as possible about what kind of data they have without needing to manually check - but if it means trading off complexity in the underlying type implementation that's obviously a problem.

I also notice that in `guiSpecToColumnProps`, we get TS errors because the `isMulti = spec.type === "multicolumn";` isn't enough to narrow the type.

xinaesthete:
@coderabbitai Interface 'SingleColumnSelection<T>' incorrectly extends interface 'BaseColumnSelection'.
  Types of property 'columnTypeFilter' are incompatible.
    Type 'T | T[] | undefined' is not assignable to type '"number" | DataType | DataType[] | undefined'.
      Type 'T[]' is not assignable to type '"number" | DataType | DataType[] | undefined'.
        Type 'T[]' is not assignable to type 'DataType[]'.
          Type 'T' is not assignable to type 'DataType'.
            Type '"number" | DataType' is not assignable to type 'DataType'.
              Type '"number"' is not assignable to type 'DataType'.ts(2430)

xinaesthete:
@coderabbitai in retrospect I think I'm going to keep the changes smaller in the scope of this PR and hope to revisit with some more rationalisation in the near future. So I'd like to focus on just making sure that each tab in the column selection UI works as it should - perhaps with each one having a specific hook responsible for narrowing the incoming generic type and providing an interface to the rest of the component for accessing and setting state. We may still want to get rid of that extra `multiple?: boolean` prop, though.

xinaesthete:
@coderabbitai let's focus on this `"multiple"` thing.

Let's consider the ways we specify - as a 'user' of the component - what it is we want.

We might have a `GuiSpec<"multicolumn">`, or something in `BaseChart.types` with a param `"_multi_column:number"`.

Either way, we end up in `ColumnSelectionComponent` with some `ColumnSelectionProps` adapted from the original props.

I think that the problem isn't so much the existence of the `multiple` prop - but that it is optional and not consistently applied. If it was non-optional & we just made sure that whenever we adapt other types of config into this type, we always had to set it, that would make the internals of `ColumnSelectionComponent` have a much simpler job.

It may be that we later decide not to have things like `"_multi_column:number"`, in favour of just having two properties. I'd be tempted to make that change now, but I think I'm satisfied that getting rid of `multiple` is a step away from the direction I want to take.

xinaesthete:
@coderabbitai what I'm hoping is that since `ColumnSelectionProps` can know whether it's `multiple`, that means that it can also type the `current_value` and `setSelectedColumn` accordingly:

```ts
export type ColumnSelectionProps<T extends CTypes, M extends boolean,
    V = M extends true ? FieldSpecs : FieldName | MultiColumnQuery> = { //should we have a SingleColumnQuery?
        type?: T; //wary of using 'type' as a name - not reserved, but could be confusing. also wary of optional type
        multiple: M; //also interacts with type "_multi..."; in future, prefer separate props for multiple selection vs datatype
        setSelectedColumn: (column: V) => void;
        current_value?: V;
        placeholder?: string;
        exclude?: string[];
        dataStore?: DataStore;
    };
```

The annoying thing is that this seems to leak into me needing more complex type signatures in places that use this & related types. How should I avoid that?

src/charts/dialogs/utils/ParamsSettingGui.tsx
xinaesthete, line 1: @coderabbitai note my comment about compatible data-type information not being passed here... what shall we do so this follows smoothly?

src/react/components/ColumnDropdownComponent.tsx
```diff
@@ -74,22 +79,24 @@ const ColumnDropdownComponent = observer(<T extends CTypes, M extends boolean>(g
...
                             setSelectedColumn(value.map(v => v.field));
                         } else {
-                            if (isArray(value)) throw new Error("Expected single value - and really didn't expect to get here");
-                            //@ts-expect-error haven't quite managed to infer the setSelectedColumn type here...
+                            if (isArray(value)) throw new Error("Unexpected array - this should be unreachable");
+                            //@ts-expect-error we could use a narrower type for the setter we use here
                             setSelectedColumn(value.field);
                         }
                     }}
+                    //@ts-expect-error wtf is going on, why would an individual option be an array?
```
xinaesthete, line 99: @coderabbitai can you tell what's causing the type that the Autocomplete component is passing to `getOptionLabel` & `renderOption` to possibly be an array?

xinaesthete: This seems to be true - looking back at the `DropdownAutocompleteComponent` code I have a similar issue & deal with it.

xinaesthete: I only use `@ts-expect-error` for things I want to flag as wrong and WIP (although they do hang around like a bad `todo` smell at times)... I digress; could you suggest some concise changes along those lines?

xinaesthete: https://github.com/coderabbitai I only use @ts-expect-error for things I want to flag as wrong and WIP (although they do hang around like a bad todo smell at times)... I digress; could you suggest some concise changes along those lines?

xinaesthete: @coderabbitai I only use `@ts-expect-error` for things I want to flag as wrong and WIP (although they do hang around like a bad `todo` smell at times)... I digress; could you suggest some concise changes along those lines?

xinaesthete: @coderabbitai the learning shouldn't be that the codebase uses `@ts-expect-error` to workaround, but to flag where solutions are being sought - the objective is to squash all of those before merging to a stable branch.

xinaesthete:
@coderabbitai well... efforts to keep changes to reasonably atomic PRs have not been possible in this case... but I think I'm finally up to a point of not having any particularly glaring bugs with this implementation... what I need to do now (well, after the weekend) is some final cleanup & get this merge back to dev ASAP... so I guess it would be good if you could go over any loose ends you can spot (`ts-expect-error` is definitely still a thing), perhaps some redundancy we could get rid of... general strategy for having a normal working dev branch again & keeping the merge commits as clear as possible...

PR #148, opened 2025-03-18
Popout Font and Chart fix

nawabfurquan:


src/utilities/Popout.ts
nawabfurquan, line 1: @coderabbitai There is a problem with the popout functionality. When we first popout a chart, the chart's inner content doesn't resize based on the size of the new popout window. The axes don't seem to expand, they look the same size as they were in the main window. But when I go back to main window and come back to the popout window, the chart's content appears to be full screen and gets adapted to the popout window perfectly. Note that, only the chart's content has this issue, the title bar adjusts perfectly fine. One more thing to note is that, this behaviour is not consistent around all browsers, the above mentioned issue is in Chrome and Firefox, Safari seems to work perfectly. Help me fix this issue.

nawabfurquan: @coderabbitai I am afraid, that doesn't seem to solve the problem. Let me know if you need images to understand the issue.

nawabfurquan: @coderabbitai
image
The above image is how it looks like in popout window

image
This image is how it looks like in main window

nawabfurquan: @coderabbitai This again doesn't seem to fix the issue on Google chrome at all, it looks exactly it did earlier. In case of firefox, it does seem to adjust it's size to the window but there seems to be a slight delay to it.

nawabfurquan: @coderabbitai That also does not fix the problem. I am getting this error in chrome for 2d scatter plot chart: WGLChart: bad _getContentDimensions()
setSize	@	WGLChart.js:134
popoutChart	@	Popout.ts:136
_popOutChart	@	ChartManager.js:2357
cm._popOutChart	@	InterProcessCommunication.ts:89
func	@	ChartManager.js:2115
(anonymous)	@	BaseChart.ts:377``

And a different error for a scatterplot.
But this error only appear in chrome when we popout the chart, it doesn't appear in firefox.

xinaesthete:
This seems mostly good, but I still see an issue when I first popout, then go fullscreen in Safari.

xinaesthete:
@coderabbitai can you create an issue linked to this, mentioning that there is a specific issue on Safari that seems to apply when you pop-out a chart to a separate window, and then make it fullscreen. For the older vanilla-js charts, it often ends up with axes that only occupy the width of the popped-out window, and then gl content is proportioned correctly, but sometimes (randomly, presumably relating to some async indetermism) clipped to that width.

There is also a separate issue, applying to all browsers, where the react-based viv chart rendered in this way doesn't have proper mouse events attached, even though it is correct with either popout or fullscreen, the combination has an issue.

PR #149, opened 2025-03-18
Dialog Drag Issue Fix

nawabfurquan:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved draggable element behavior with enhanced boundary checks to ensure elements remain visible within the window during dragging.

### DIFF
--- a/src/utilities/Elements.js
+++ b/src/utilities/Elements.js
@@ -485,10 +485,17 @@ function makeDraggable(el, config = {}) {
         // If it exceeds the window, reset it to 0
         if (nt < 0) nt = 0;
 
-        // if it is (nearly) off the left-hand side of the window, reset it to 0
-        const { right } = el.getBoundingClientRect();
+        // if it is (nearly) off the side of the window, keep it within 100px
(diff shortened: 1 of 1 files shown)


nawabfurquan:
@coderabbitai review

nawabfurquan:
@coderabbitai I would like you to explain this codebase to me as I am a fairly new developer to this project. Explain me the architecture and help me understand how the application works.

xinaesthete:
There is an issue where if the mouse-up event happens while the mouse is outside of the window bounds, it will be stuck dragging (I think you have to refresh the page!).

Also I'm going to change it so that if it goes off the left-hand side, it will push it back inside.

xinaesthete:
I've now made it so that on each edge, if there will be less than 100px visible on that edge after dragging, it will snap back.

This allows users to push things out of the way, without risking items getting totally lost.

It is still possible if the window resizes, that the element will end up somewhere to the bottom and/or right.... it is still possible to scroll to it in that case, although ideally we'd listen to resize events and snap back when needed (also intersects with popout behaviour...)

PR #150, opened 2025-03-19
Autocomplete fixes and enhancements

nawabfurquan:
## Summary by CodeRabbit

- **New Features**
  - Multi-select dropdowns now include a “Select All” option for easier full selection.
  - Dropdown interactions have been enhanced with improved open/close behavior, including automatic closure when the element is out of view.
  - Tag displays are now optimized by limiting visible tags and summarizing additional selections.
  - Overall, these improvements offer a smoother and more intuitive selection experience across dialogs.

### DIFF
--- a/src/react/components/ColumnDropdownComponent.tsx
+++ b/src/react/components/ColumnDropdownComponent.tsx
@@ -2,7 +2,7 @@ import { columnMatchesType, inferGenericColumnSelectionProps } from "@/lib/colum
 import type { ColumnSelectionProps, CTypes } from "@/lib/columnTypeHelpers";
 import { useDataStore } from "../context";
 import type { DataColumn, DataType } from "@/charts/charts";
-import { useEffect, useMemo, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import { observer } from "mobx-react-lite";
 import { Autocomplete, Box, Checkbox, Chip, Divider, FormControlLabel, Paper } from "@mui/material";
 import { isArray } from "@/lib/utils";
@@ -10,6 +10,7 @@ import { TextFieldExtended } from "./TextFieldExtended";
 import Grid from '@mui/material/Grid2';
 import CheckBoxOutlineBlankIcon from "@mui/icons-material/CheckBoxOutlineBlank";
 import CheckBoxIcon from "@mui/icons-material/CheckBox";
+import { useCloseOnIntersection } from "../hooks";
 
 const icon = <CheckBoxOutlineBlankIcon fontSize="small" />;
 const checkedIcon = <CheckBoxIcon fontSize="small" />;
@@ -89,6 +90,8 @@ const ColumnDropdownComponent = observer(<T extends CTypes, M extends boolean>(g
     //todo fix this type error
     //@ts-expect-error
     const [selectAll, setSelectAll] = useState(isMultiType ? columns.length === value?.length : false);
+    const [open, setOpen] = useState(false);
+    const ref = useRef<HTMLInputElement>(null);
 
     useEffect(() => {
         if (isMultiType) {
@@ -97,6 +100,8 @@ const ColumnDropdownComponent = observer(<T extends CTypes, M extends boolean>(g
             setSelectAll(value?.length === columns.length);
         }
     }, [value, columns.length, isMultiType]);
+
+    useCloseOnIntersection(ref, () => setOpen(false));
     // should we be using a `GuiSpec<"dropdown"> | GuiSpec<"multidropdown">` here?
     // we decided against - we have some extra adornments we add and we don't want to have to deal with that in GuiSpec
     // (at least, not yet - might be tempted but there's a serious risk of types getting further out of hand)
@@ -127,6 +132,10 @@ const ColumnDropdownComponent = observer(<T extends CTypes, M extends boolean>(g
                     multiple={isMultiType}
                     value={value as ColumnValue}
                     disableCloseOnSelect={isMultiType}
+                    open={open}
+                    onOpen={() => setOpen(true)}
+                    onClose={() => setOpen(false)}
+                    ref={ref}
                     onChange={(_, value) => {
                         //! now that we say `value as ColumnValue`, MUI thinks the value for onChange is not an array - which it could be.
                         // *sigh*. time to move on.
--- a/src/react/components/ColumnDropdownComponent.tsx
+++ b/src/react/components/ColumnDropdownComponent.tsx
@@ -87,17 +87,13 @@ const useColumnDropdownValue = <T extends CTypes, M extends boolean>(gProps: Col
  */
 const ColumnDropdownComponent = observer(<T extends CTypes, M extends boolean>(gProps: ColumnSelectionProps<T, M>) => {
     const { placeholder, isMultiType, columns, value, setValue } = useColumnDropdownValue(gProps);
-    //todo fix this type error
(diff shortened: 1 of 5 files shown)


nawabfurquan:
@coderabbitai We have an issue with the `Autocomplete` component of MUI which is used in `DropdownAutocompleteComponent`, `ColumnDropdownComponent` and `TextComponent`, when it's dropdown is open and we scroll over it's container, the dropdown also scrolls with it. I am attaching an image for more clarity. Help me fix this issue.
image

nawabfurquan:
@coderabbitai This too doesn't seem to solve the problem. It still scrolls with content and it's size is way smaller. Below is the image for more description. The code that I have after changing is: "export const CustomPopper = forwardRef<HTMLDivElement, PopperProps>((props, ref) => {
    return (
      <Popper
        {...props}
        ref={ref}
        modifiers={[
          {
            name: 'preventOverflow',
            options: {
              altAxis: true,
              altBoundary: true,
              tether: true,
              rootBoundary: 'document',
              padding: 8,
            },
          },
          {
            name: 'flip',
            options: {
              altBoundary: true,
              rootBoundary: 'document',
              padding: 8,
            },
          }
        ]}
        style={{ zIndex: 1500 }} // Ensure it's above dialogs
      />
    );
  });
export const DropdownAutocompleteComponent = observer(({
    props//: propsAmbiguous,
}: { props: DropdownSpec }) => {
    // todo review 'virtualization' for large lists
    const id = useId();
    // const props = inferDropdownType(propsAmbiguous);
    const multiple = isMultidropdown(props);
    // the props.values may be a tuple of [valueObjectArray, textKey, valueKey], or an array of length 1 - [string[]]
    if (!props.values) throw new Error("DropdownAutocompleteComponent requires props.values");
    //todo handle multitext / tags properly.

    const {options, single, label, val, okOption} = useDropdownOptions(props);

    type Option = typeof options[number]; // hopefully we can improve `{ original: any }`
    // still not entirely sure about the type for onChange...
    type OVal = Option | Option[] | (Option | Option[])[] | null;
    return (
        <>
            <MLabel htmlFor={id} props={props} />
            <Autocomplete
                className="w-full"
                multiple={multiple}
                size="small"
                id={id}
                options={options}
                limitTags={5}
                disableCloseOnSelect={multiple}
                getOptionLabel={label}
                value={okOption.filter((a) => a !== undefined)}
                onChange={action((_, value: OVal) => {
                    //added type annotation above because mobx seems to fluff the inference to `never`
                    if (value === null) return;
                    if (isArray(value)) {
                        if (!multiple) throw new Error("shouldn't get an array here");
                        const vals = value.map(val);
                        props.current_value = vals;
                        props.func?.(vals);
                        return;
                    }
                    if (multiple) throw new Error("shouldn't get a single value here");
                    const v = val(value);
                    props.current_value = v;
                    if (props.func) props.func(v);
                })}
                isOptionEqualToValue={(option, value) => (single(option).original === single(value).original)}
                renderOption={(props, option, { selected }) => {
                    if (!option) return null; //FFS let's switch to Rust or something
                    // we could potentially render something richer here - like color swatches or icons
                    // if we had a richer sense of the data in context
                    // ^^ e.g. if we had a `GuiSpec<'category'>` it could have it's own internal logic for
                    // managing internal state, and also use a richer component here.
                    const { key, ...optionProps } = props as typeof props & {
                        key: string;
                    }; //questionable mui types?
                    if (multiple)
                        return (
                            <li key={key} {...optionProps}>
                                <Checkbox
                                    icon={icon}
                                    checkedIcon={checkedIcon}
                                    style={{ marginRight: 8 }}
                                    checked={selected}
                                />
                                {label(option)}
                            </li>
                        );
                    return (
                        <li key={key} {...optionProps}>
                            {label(option)}
                        </li>
                    );
                }}
                renderTags={(tagValue, getTagProps) => {
                    //seems to be a material-ui bug with not properly handling key / props...
                    //https://stackoverflow.com/questions/75818761/material-ui-autocomplete-warning-a-props-object-containing-a-key-prop-is-be
                    return tagValue.map((option, index) => !option ? null : (
                        <Chip
                            {...getTagProps({ index })}
                            key={val(option)}
                            label={label(option)}
                        />
                    ));
                }}
                renderInput={(params) => (
                    <TextField
                        {...params}
                    // label="Checkboxes"
                    // placeholder={props.label}
                    />
                )}
                PopperComponent={CustomPopper}
            />
        </>
    );
});"

image

nawabfurquan:
@coderabbitai This hides the whole component with the dropdown on top of it. Below is the image:

image

It also scrolls with the content:

image

nawabfurquan:
@coderabbitai No, this doesn't solve the issue. It still scrolls with the content
image

nawabfurquan:
@coderabbitai I am afraid, that fix also didn't help. Is there no simple solution for it?

src/react/components/ColumnDropdownComponent.tsx
```diff
@@ -87,12 +87,37 @@ const useColumnDropdownValue = <T extends CTypes, M extends boolean>(gProps: Col
...
     // should we be using a `GuiSpec<"dropdown"> | GuiSpec<"multidropdown">` here?
     // we decided against - we have some extra adornments we add and we don't want to have to deal with that in GuiSpec
     // (at least, not yet - might be tempted but there's a serious risk of types getting further out of hand)
     type ColumnOption = DataColumn<DataType>;
     type ColumnValue = M extends true ? ColumnOption[] : ColumnOption | null;
+
+    const handleSelectAll = () => {
+        if (!isMultiType) throw new Error("expected multitype here");
+        if (selectAll) {
+            //todo remove this ts-expect-error
```
nawabfurquan, line 110: @coderabbitai Help me solve the type issues in line 112 and line 117

src/react/components/ColumnDropdownComponent.tsx
```diff
@@ -87,12 +87,35 @@ const useColumnDropdownValue = <T extends CTypes, M extends boolean>(gProps: Col
  */
 const ColumnDropdownComponent = observer(<T extends CTypes, M extends boolean>(gProps: ColumnSelectionProps<T, M>) => {
     const { placeholder, isMultiType, columns, value, setValue } = useColumnDropdownValue(gProps);
-    
+    const [selectAll, setSelectAll] = useState(isMultiType && Array.isArray(value) ? columns.length === value?.length : false);
```
xinaesthete, line 90: btw we have an `isArray()` utility function that also works as a type-predicate, which the native one doesn't... but it isn't helping as much as I'd like to let TS know whether our type `M` matches the value being an array...

src/react/hooks.ts
```diff
@@ -406,3 +407,25 @@ export function useRangeDimension2D() {
...
+
+
+export const useCloseOnIntersection = (ref: React.RefObject<HTMLElement>, onClose: () => void) => {
+    useEffect(() => {
+        if (!ref.current) return;
+        // Observer to observe the an element and close it when it intersects the parent element (most likely while scrolling)
+        const observer = new IntersectionObserver((entries) => {
+            entries.forEach((entry) => {
+                // If less than half of the element is intersecting, call onClose
+                if (entry.intersectionRatio < 0.5) {
```
xinaesthete, line 419: we might change this so that the entire element needs to be non-intersecting or something... I've noticed some slightly unexpected behaviour e.g. if you have a dialog which is only partially within the screen, and then you click to bring up the dropdown and it automatically hides straight away...

PR #157, opened 2025-03-26
Fix problems with popout charts in gridstack layout

martinSergeant:
**Problems**
When charts were added back to a gridstack layout they reverted to absolute positioning
In gridstack there were no margins and the charts were slightly overlapping

**Fixes**
Overhauled gridstack manager adding outer and inner containers to ensure it works correctly e.g. correct margins
Also ensured charts can be added and removed from the gridstack  without errors  - however it has not been extensively tested
Ensure layout can be changed even when some charts are popped out

### DIFF
--- a/src/charts/GridstackManager.ts
+++ b/src/charts/GridstackManager.ts
@@ -278,8 +278,8 @@ export default class GridStackManager {
                 locked = !locked;
                 lockIcon.classList.toggle("fa-lock", locked);
                 lockIcon.classList.toggle("fa-unlock", !locked);
-                grid.update(div, { locked });
-                div.classList.toggle("gridLock", locked);
+                grid.update(outer, { locked });
+                outer.classList.toggle("gridLock", locked);
             });
             return lockButton;
         }
--- a/src/charts/GridstackManager.ts
+++ b/src/charts/GridstackManager.ts
@@ -73,12 +73,34 @@ export default class GridStackManager {
                 ds.name,
                 "fas fa-compress-arrows-alt",
                 "compact layout",
-                () => grid.compact(),
+                () => this.compact(ds),
             );
             this.grids.set(ds, { grid: grid, charts: new Set(), icon: i });
         }
 
         return this.grids.get(ds);
+    }
+
+    //when the grid calls compact the size and move observers are not called
+    //therefore need to do it manually
+    compact(ds: DataSource) {
+        const gi = this.grids.get(ds);
+        if (!gi) {
+            console.error(`no gridstack for ${ds.name}`);
+            return;
+        }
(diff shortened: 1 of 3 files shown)


src/charts/GridstackManager.ts
```diff
@@ -204,22 +270,32 @@ export default class GridStackManager {
...
             };
             return remove;
         }
+
+        div.gridstackPopoutCallback = () => {
+            chart.removeLayout?.();
+        };
+
+        //the below was leading subtle errors
+        //not sure why you need to alter changeBaseDocument?
```
xinaesthete, line 290: hmmm... at various points I lost track a bit of various versions of `changeBaseDocument` logic - for the react stuff,  `useOuterContainer` is now used for related things - it may be that some water could be un-muddied.

xinaesthete:
Tooltips are no-longer unclipped by the chart div:

Screenshot 2025-03-28 at 12 18 43

xinaesthete:
An existing bug we have (not changed here I think) is that switching layout a few times often results in some charts shifting around.

martinSergeant:
>Tooltips are no-longer unclipped by the chart div:

This has always been the case and there is no easy fix. Making the chart div css overlow:visible does fix the issue but causes many other problems. The fundamental problem with microtip is that it attached to an element and not the document Best to replace microtip with another library e.g. material ui , which is going to happen anyway

>Locking position of a gridstack element no longer prevents it being moved e.g. by 'compact layout' action, or other charts trying to occupy the space.

Fixed

> An existing bug we have (not changed here I think) is that switching layout a few times often results in some charts shifting around.

Yes,  but his only happens with complex layouts when you swap between grid-stack and absolute many times - maybe not a priority if there is no quick fix

xinaesthete:
> > Tooltips are no-longer unclipped by the chart div:
> 
> This has always been the case and there is no easy fix.

Yes I realised after first commenting that it was an existing issue.

> 
> > Locking position of a gridstack element no longer prevents it being moved e.g. by 'compact layout' action, or other charts trying to occupy the space.
> 
> Fixed

Thanks.

> 
> > An existing bug we have (not changed here I think) is that switching layout a few times often results in some charts shifting around.
> 
> Yes, but his only happens with complex layouts when you swap between grid-stack and absolute many times - maybe not a priority if there is no quick fix

Agreed - ultimately I think we may want something else for more complex layout anyway.

I've rebased-in some other changes from dev - including charts having a `deferredInit` method for things that should happen once they have some notional "settled state" during initialisation... I guess that could let us avoid the 'no data for row chart' thing, if we're concerned. Added a bit of a comment to the code on that.

PR #158, opened 2025-03-27
Updata anndata and mudata conversions

martinSergeant:
Contains a two fixes to the ui (which may now be obsolete)
Also updated the conversions from anndata and mudata objects

## Summary by CodeRabbit

- **New Features**
  - Introduced an option for processing large datasets in chunks, helping to improve performance and reduce memory usage.
  - Updated genome browser tracks to automatically generate custom track names for better file organization.
  
- **Bug Fixes**
  - Enhanced the handling of genomic data retrieval, resulting in more reliable and error-free loading.
  - Optimized the merging of multi-modal data to ensure consistent display of results.

### DIFF
--- a/python/mdvtools/conversions.py
+++ b/python/mdvtools/conversions.py
@@ -15,7 +15,8 @@ def convert_scanpy_to_mdv(
     scanpy_object: AnnData, 
     max_dims: int = 3, 
     delete_existing: bool = False, 
-    label: str = ""
+    label: str = "",
+    chunk_data: bool = False
 ) -> MDVProject:
     """
     Convert a Scanpy AnnData object to MDV (Multi-Dimensional Viewer) format.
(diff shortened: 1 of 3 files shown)


xinaesthete:
I think we'll probably revert the `ChartManager` changes.

I might just merge the other bits with fairly minimal testing - is there an example of something that demonstrates the difference in behaviour? Maybe ping me separately if it's non-public data.

xinaesthete:
Must admit going a bit misty-eyed reading the old simpler `_decorateColumnMethods` code...

xinaesthete:
Also as mentioned, would ideally like 'gene column' (rows_as_columns_link.name_column) to be allowed to be unique, will have to get back to that...

I've just rebased `dev` into a local copy of this branch, which didn't have any conflicts apart from ChartManager and I think will lead to a slightly cleaner history than merging, but I'm slightly hesitant to push onto your branch in case it messes something up for you.

xinaesthete:
For ref, I tested it against our `ilc_viz_ready_revised.h5ad`. Would be good to make some tests at some point.

@coderabbitai review

PR #166, opened 2025-04-04
useScatterRadius hook

xinaesthete:
@coderabbitai the intention here is to have more sane control of point size, as per comments on `useScatterRadius()`... but I've realised I recently introduced a bug meaning that `useScatterplotLayer()` is not reacting to changes in the mobx `config`. I'm using `autorun` in a `useEffect` and now it does re-render when I tweak those settings - but other related config things are not working correctly since I moved the call to `useScatterplotLayer()` inside `useCreateSpatialAnnotationState()`.

### DIFF
--- a/src/react/components/DeckScatterComponent.tsx
+++ b/src/react/components/DeckScatterComponent.tsx
@@ -14,6 +14,7 @@ import { Axis, Scale } from "@visx/visx";
 import "../../charts/css/charts.css";
 import { SpatialAnnotationProvider, useSpatialLayers } from "../spatial_context";
 import SelectionOverlay from "./SelectionOverlay";
+import { useScatterRadius } from "../scatter_state";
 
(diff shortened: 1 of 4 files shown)


xinaesthete:
@coderabbitai the intention here is to have more sane control of point size, as per comments on useScatterRadius()... but I've realised I recently introduced a bug meaning that useScatterplotLayer() is not reacting to changes in the mobx config. I'm using autorun in a useEffect and now it does re-render when I tweak those settings - but other related config things are not working correctly since I moved the call to useScatterplotLayer() inside useCreateSpatialAnnotationState().

xinaesthete:
@coderabbitai the issue is that I want to have a context with a shared instance of the same `scatterProps` - it causes other issues. I suppose if I had a `mobxAutorun` at the outer-layer of `useScatterplotLayer()` that would probably work - not sure what extra overhead this is adding; when I tweak radius it can feel quite noticeably laggy.

PR #169, opened 2025-04-09
Md 476 viv histogram widgets

xinaesthete:
@coderabbitai I'm doing something very wrong with my state management the way `ChannelHistogram` is used for colour channel limit control - I thought it was working reasonably ok at some point recently (before I went to bed), but it just goes into an infinite loop now. The original use seems ok.


xinaesthete:
@coderabbitai review, and suggest saner state management in `Histogram` / `ChannelHistogram`...

xinaesthete:
@coderabbitai review

xinaesthete:
@coderabbitai, I haven't looked in detail, but what you suggest sounds reasonable and like it should work... I wonder if we could simplifying our state management more, instead - this seems to be layering on more levels of complexity to work around inherent issues with the design.

xinaesthete:
@coderabbitai I definitely want the state of the rest of the application to update in realtime during the interaction, not only on brush interaction end.

xinaesthete:
@coderabbitai I'm not sure if the `observer()` wrapper is necessary - does it look like we're actually using mobx in any meaningful way with `Histogram`?

xinaesthete:
@coderabbitai can you author a commit with these changes?

xinaesthete:
@coderabbitai a few issue in there... one is that while you were technically right to suggest `<rect>`s for histogram... I didn't really want that, and I should probably have named the component differently, or I might introduce a prop for choosing whether to use line or bars. I'm also getting a lot of warnings about `useEffect` for the d3 brush not specifying all of its dependencies... but we definitely don't want to be re-running that and setting up listeners again constantly... There was also a minor thing with using array index as a key as well.

xinaesthete:
@coderabbitai can you highlight the specific relevant state-management changes rather than printing the whole code for the components? There are a few other changes to visual styling as well as code formatting and how the function signatures are written etc that I don't want, so I'd like you to concisely summarise the important parts about state.

xinaesthete:
@coderabbitai actually, maybe we should be using visx to help with this d3/react bridging?

PR #171, opened 2025-04-14
Md 235 persist selection

xinaesthete:
The `FeatureCollection` used to represent selection state in scatterplots and spatial/viv chart is persisted via mobx `config`.

A serious performance issue was noticed, which is mostly mitigated by making sure that we use `toJS(coords)` before passing them on to the point-in-polygon `filterPoly` method - although there still appears to be a significant overhead, which suggests that we need to be extremely careful with how `mobx` is used.

The performance of `filterPoly` may in future be improved with changes to how we compute `Dimension`, particularly in cases where other dimensions already filter most of the samples.

@coderabbitai review

## Summary by CodeRabbit

- **Refactor**
	- Streamlined scatter plot configuration settings to simplify customization.
	- Enhanced state management for interactive map features, resulting in smoother performance during user interactions.

### DIFF
--- a/src/react/components/VivMDVReact.tsx
+++ b/src/react/components/VivMDVReact.tsx
@@ -1,4 +1,4 @@
-import BaseChart, { type BaseConfig } from "../../charts/BaseChart";
+import BaseChart from "../../charts/BaseChart";
 import { BaseReactChart } from "./BaseReactChart";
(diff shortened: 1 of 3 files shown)


xinaesthete:
@coderabbitai review

PR #180, opened 2025-04-22
Md 485 tooltips for deck.gl scatterplot, hook for fields in config

xinaesthete:
Both deck scatterplots & react-viv charts now use common code for tooltips, and associated settings dialog entries.

`useFieldSpecs(specs)` hook has been added, which is generally useful any time a chart config object has some arbitrary property representing column data, rather than requiring that fields either be listed in `param` or have `configEntriesUsingColumns` (which is only useful for non-nested properties) etc.

New tooltips accept multi-column, so that users can get a bit more useful insight from them.

## Summary by CodeRabbit

- **New Features**
  - Enhanced tooltip functionality in scatterplots to support displaying multiple columns simultaneously.
  - Tooltip settings now provide a multiselect interface for choosing multiple columns to display in tooltips.

- **Bug Fixes**
  - Tooltips now correctly display all selected columns, resolving previous issues with array handling.

- **Refactor**
  - Consolidated tooltip settings UI into a reusable utility for consistency across components.
  - Improved hooks for loading and managing multiple data columns dynamically.

- **Chores**
  - Cleaned up unused imports and improved dependency tracking in memoized React hooks.

### DIFF
--- a/src/charts/ChartManager.js
+++ b/src/charts/ChartManager.js
@@ -948,7 +948,13 @@ export class ChartManager {
         }
         if (config.tooltip) {
             if (config.tooltip.column) {
-                set.add(config.tooltip.column);
+                if (Array.isArray(config.tooltip.column)) {
+                    for (const i of config.tooltip.column) {
+                        set.add(i);
+                    }
+                } else {
+                    set.add(config.tooltip.column);
+                }
             }
         }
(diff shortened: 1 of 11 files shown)


xinaesthete:
@coderabbitai review

PR #189, opened 2025-04-28
Document Title fix

nawabfurquan:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling during chart view saving to log errors and manage them more gracefully.
	- Updated project data fetching to use a relative URL, ensuring compatibility across different environments.

### DIFF
--- a/src/charts/ChartManager.js
+++ b/src/charts/ChartManager.js
@@ -660,7 +660,10 @@ export class ChartManager {
                         const state = this.getState();
                         if (!state.view?.viewImage) {
                             // todo: update the error handler function in future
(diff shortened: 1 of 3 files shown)


xinaesthete:
It seems like there are more glitches with various aspects of view saving than I'd realised, potentially in at least some cases related to viv charts... I don't think there's any kind of regression here - but also I haven't really managed to meaningfully tell if the different error handling is helping / how it's working... so it perhaps could be merged without too much risk, but I think I'm going to try to do some more testing in it and potentially try to push some changes relating to viv bugs here...

PR #206, opened 2025-05-23
Refactor Import Project

nawabfurquan:


nawabfurquan:
@coderabbitai review

xinaesthete:
I haven't thought a great deal about the specific changes to import logic - although I have made it so that there is also some similar validation done in the front-end before we start uploading, so hopefully that will discourage users from doing inappropriate things.

PR #209, opened 2025-05-28
Taurus frontend changes

martinSergeant:
Changes for the Taurus data 

* In PanZoom.js (probably be deprecated soon) the base css img settings were interfering with pan/zooming. Fixes to prevent this
* Only datasources present in the current view could be  used to create a new view - this was changed so that any datasource could be used
* added an option so that filters in a selection dialogue are not removed on reset all. This used to be default but was then removed (which makes sense) . Now is optional  can be used for  a view of e.g. a single cell type and will not reset to all cells. Other dialogues for querying can be added with default behaviour

## Summary by CodeRabbit

- **New Features**
  - Added an option to prevent filters from being cleared when using "Reset All" in selection dialogs. Users can enable this via a new checkbox in the settings.
- **Improvements**
  - Enhanced the way data sources are managed and selected when adding new views, providing a more consistent experience.
  - Improved image zoom and pan functionality for smoother resizing and better aspect ratio handling.
  - Updated filter handling to respect the new "no clear filters" configuration, improving filter persistence behavior.
  - Fixed filter property naming for better consistency and reliability.

### DIFF
--- a/src/charts/dialogs/AddViewDialog.tsx
+++ b/src/charts/dialogs/AddViewDialog.tsx
@@ -18,14 +18,19 @@ const AddViewDialogComponent = (props: {
     const [viewName, setViewName] = useState("");
     const [isCloneView, setIsCloneView] = useState(false);
     const [checkedDs, setCheckedDs] = useState<{ [name: string]: boolean }>({});
+    //must be a better way than using the window object
     const { viewManager, viewData } = window.mdv.chartManager;
-    const dataSources = useMemo(() => Object.keys(viewData.dataSources), [viewData.dataSources]);
+    //we want a choice of all datasources (not just ones on the current view)
+    //not sure how to get all the datasources available - this is just a temporary hack
+    //if we don't use useMemo get constant re-rendering ?
+    const dataSources = useMemo(()=>viewManager.cm.dataSources.map(x=>x.name),[viewManager.cm])
+    const currentDataSources = useMemo(() => Object.keys(viewData.dataSources), [viewData.dataSources]);
 
     useEffect(() => {
         // Initialising the checkedDs with the data sources inside the view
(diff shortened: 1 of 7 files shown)


src/charts/dialogs/AddViewDialog.tsx
```diff
@@ -18,22 +18,27 @@ const AddViewDialogComponent = (props: {
     const [viewName, setViewName] = useState("");
     const [isCloneView, setIsCloneView] = useState(false);
     const [checkedDs, setCheckedDs] = useState<{ [name: string]: boolean }>({});
+    //must be a better way than using the window object
     const { viewManager, viewData } = window.mdv.chartManager;
-    const dataSources = useMemo(() => Object.keys(viewData.dataSources), [viewData.dataSources]);
+    //we want a choice of all datasources (not just ones on the current view)
+    //not sure how to get all the datasources available - this is just a temporary hack
+    //if we don't use useMemo get constant re-rendering ?
+    const dataSources = useMemo(()=>viewManager.getAllDataSources(),[viewManager])
```
xinaesthete, line 26: Yes, if you call some function which returns a new array, that will cause react to re-render which can be avoided with `useMemo`. 

In future, the react compiler should help with this type of thing, although given that you get different arrays every time you call this, it probably wouldn't cause it to behave differently (there would be horrendous bugs in general if it did).

We could have a more mobx way of doing things with some `observable` / `computed` properties on `ViewManager` and then if this component was wrapped in `observer` it should re-render precisely when those properties changed.

I probably wouldn't have a method `getAllDataSources()` which returns `string[]` - I'd prefer to have the objects themselves and then refer to `name` later... actually there is already a trivial `useDataSources()` hook which just wraps returning `cm.dataSources` - I think I'm going to change this to use that.

PR #217, opened 2025-06-02
Pip install

martinSergeant:
* relaxed versions of dependencies for pip installable version. Now works with python 3.10+, 3.9 can be supported but would require changes in typing
* added a light server for the pip installable version

## Summary by CodeRabbit

- **New Features**
  - Introduced a lightweight server for serving project data, metadata, and files with improved route handling and security headers.
  - Added utility functions for enhanced file serving and HTTP response management, including support for HTTP range requests and security headers.
  - Updated project serve method to open the browser automatically and use the new lightweight server.

- **Bug Fixes**
  - Improved handling of bed and bigWig file sourcing based on configuration, ensuring correct file selection for different pipeline setups.

- **Documentation**
  - Revised documentation for clarity, improved grammar, updated example paths, instructions, and corrected default server port.

- **Chores**
  - Updated project dependencies, minimum Python version, and build configuration for broader compatibility and easier installation.


xinaesthete:
I do agree on decoupling and trying to avoid keeping it bloated, I just think we need a bit of a strategy for not showing bits of the frontend that wouldn't work if an unsuspecting user interacts with them. I think charts themselves could potentially be less internally coupled so that they could be used in notebooks etc as well.

There'd be a lot to be said for more formally segregating stuff related to our catalog/dashboard etc. The other thing is stuff like the routes for adding data to projects etc, which I think we need to not show in the frontend if it's being served in a way that won't have corresponding endpoints.

xinaesthete:
As discussed, my only material concern in terms of merging was that end-users get served UI (in particular, for file upload) without support for the corresponding endpoints. I agree that we should support lighter-weight builds, we'll need to reach a better understanding of how that should be organised.

I think the simplest way around this in the short-term will be for the server to add something which opts-in to certain features by adding flags to `state.json` and the front-end hides corresponding functionality as we already do with chat. That way, nothing needs to change in what you do here, and we progressively-enhance with somewhat better modularisation.

So I'm happy to merge this and plan to make changes soon such that it doesn't require any action to satisfy my niggle, but there are a couple of minor conflicts now which need tidying.

xinaesthete:
The type errors pointed out in CI are indicating breaking change to the signature of the `MDVProject.serve` method which totally breaks our server.

PR #219, opened 2025-06-03
changed the background and text color of tooltip

nawabfurquan:


nawabfurquan:
@coderabbitai I need to fix the color of the tooltip which is used inside the scatter plot, which appears when we click on show tooltip in the settings. I can't find where it is, help me find it.

PR #222, opened 2025-06-05
Md 575 dotplot categories bug

xinaesthete:
There were logic errors in the way that we move between representations of `config.param`, meaning that - for instance - with Dot Plot, if the user had chosen an active query for fields, and later changed the other param for category, this would cause it to overwrite the current "concrete field names" representation stored in `chart.config`.

We now have a way of getting `activeParams` that allows more direct use of this state and getting rid of some horrible internal logic in `ParamsSettingsGui`.

I think we should have specific fields for each parameter in config objects rather than relying on `param[]` - in which case we need to consider versioning/migration of legacy configs etc.

## Summary by CodeRabbit

- **New Features**
  - Improved chart title synchronization and dynamic color updates in CategoryChart when configuration changes.
  - Added a method to access active chart parameters for better parameter management dialogs.
  - Updated authentication to use a global JWKS cache, reducing external API calls and improving performance.
  - Conditional redirection after authentication based on environment configuration.

- **Performance Optimization**
  - Implemented caching and refresh intervals for user/project data and JWKS to reduce redundant validations and database calls.
  - Reduced logging verbosity in key request paths for faster response times.

- **Bug Fixes**
  - Ensured project owner field uses a consistent name in API responses.

- **Style**
  - Unified tooltip styling across themes using CSS variables for background and text colors.
  - Updated tooltip and context menu colors to use theme variables for consistent appearance.

- **Documentation**
  - Added a comprehensive guide detailing authentication performance optimizations and recommendations.

- **Refactor**
  - Simplified parameter update logic in chart dialogs for maintainability.
  - Switched DotPlot field updates to use reactive MobX autorun instead of direct calls.

- **Chores**
  - Exported additional type definitions for chart configuration utilities.

### DIFF
--- a/src/charts/BaseChart.ts
+++ b/src/charts/BaseChart.ts
@@ -697,7 +697,7 @@ class BaseChart<T extends BaseConfig> {
     @loadColumnData
     setParams(params: FieldSpecs) {
         //! now the config will just have string[] - we want to allow things that can understand to have other objects
-        // this.config.param = params; // This line is removed/commented as ColumnQueryMapper.setActiveQuery now handles updating the live config.
+        this.config.param = params;
         // if we keep liveParams around it could be useful when serialising the chart
         // although how can we be sure that the liveParams are up to date?
         // maybe as a rule, this method is the only way to set the params...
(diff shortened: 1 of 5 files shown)


xinaesthete:
@coderabbitai I had originally based this PR on the wrong base, can you re-review with only the specific related considered?

PR #233, opened 2025-06-20
Restrict file registration to ALLOWED_FILES in project serving

bioinfbloke:
This PR was to stop a project not loading due to suspected time outs when it contains extra files other than  {'views.json', 'datasources.json', 'datafile.h5', 'state.json'}. For example, in chatMDV projects there is the scanpy file which can be called any  name (ends in .h5ad) but can be very large.

This PR ensures that only files listed in ALLOWED_FILES are registered with the backend when serving projects from the database or filesystem.


xinaesthete:
I think this is based on dev? Should we change it to merge into that - would help to see what the relevant changes are etc.

PR #239, opened 2025-07-01
Md 601 categorical columns not being properly added to project

xinaesthete:


xinaesthete:
Note @aleql @martinSergeant  there is a `test_stress_performance.py` here that is failing - written by Cursor in a very "vibe code" way for better or worse (it seems to be working out somewhat ok for this kind of test generation, but could do with proper vetting & review). The numbers of cells & genes doesn't even seem all that big. I'll try to investigate a bit more properly.

xinaesthete:
Before merging this PR, I will probably make it so that the stress-tests don't always run alongside other tests, as it seems like a waste of resources apart from anything else. We could probably consider having another CI action, maybe triggered by a `[perf-test]` in commit message or something.

FWIW the current version as of writing does still fail locally as well as in CI, I had mistakenly thought that was not the case when I last pushed. It may be because of a race-condition with accessing `datasources.json` during `convert_scanpy_to_mdv`, as opposed to resource consumption.

So this is all getting a bit more involved and it would probably best if we merge this sooner rather than later so that the original bug with categorical data is fixed.

python/mdvtools/conversions.py
```diff
@@ -137,7 +141,7 @@ def convert_scanpy_to_mdv(
         for layer,matrix in scanpy_object.layers.items():
             matrix,sparse = get_matrix(matrix)
             mdv.add_rows_as_columns_subgroup(
-                f"{label}cells", f"{label}genes", matrix, layer, sparse=sparse, chunk_data=chunk_data
+                f"{label}cells", f"{label}genes", layer, matrix, sparse=sparse, chunk_data=chunk_data
```
xinaesthete, line 144: @martinSergeant this cropped up in the process of adding tests etc - can you cast an eye over before I merge?

PR #261, opened 2025-09-02
Chart Resizer Update

nawabfurquan:


nawabfurquan:
@coderabbitai review

martinSergeant:
1.  When charts are close together, even when there is a significant gap , on the side where they are adjacent , you can only resize one chart (the one that is on top) This is an edge case and should not cause many problems.
2. Whilst resizing , the content is visible outside the chart area which looks a bit odd 
    image
3. The menu bar and the content of some charts spill out permanently   when the chart is made smaller
image

PR #267, opened 2025-09-22
Fix Socket.IO Uploads by Aligning Server/Client Paths

aleql:
This PR aims to fix the Socket.IO upload failures that were causing 500 errors when connecting. The root of the problem was a mismatch between the client and server socket paths. The client was attempting to connect using a sub-path such as /server/socket.io, while the server was only listening on the default /socket.io. This caused the upload handshake requests to hit an unhandled route, resulting in errors.

The issue was introduced in commit [78c48fa8 ](https://github.com/Taylor-CCB-Group/MDV/commit/78c48fa86746153cd71ec6c4051247eafc9f457f)(“SocketIO Import and Create Project Fix (#263)”), when the serverUrl parameter was removed and replaced with a dynamically constructed socketPath based on mdvApiRoot. This change allowed the client to adapt to sub-path deployments, but the server-side initialization in websocket.py was not updated at the same time. The result was that the client and server were no longer aligned, and uploads stopped working.

The fix was to update the server configuration so that it also respects MDV_API_ROOT when setting up the Socket.IO instance. Now the server computes the socket path dynamically: it uses /socket.io when deployed at the root, and <MDV_API_ROOT>/socket.io when deployed under a sub-path such as /server/ or /carroll/. With this change, the client and server are once again in sync, and uploads succeed regardless of deployment structure.

### DIFF
--- a/python/mdvtools/websocket.py
+++ b/python/mdvtools/websocket.py
@@ -27,7 +27,7 @@ class SocketIOContextRequest(FlaskRequest):
 def mdv_socketio(app: Flask):
     """
     Experimental and not to be trusted pending design work etc.
-    
+
     Do we have a SocketIO for the entire app and route messages internally,
     - yes probably.
     What `path` should it use? We'll need to make sure it is compatible with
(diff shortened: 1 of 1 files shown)


aleql:
PR update:
- **Reverted PR XXX**: The previous PR disabled not only WebSocket-based upload, but also the retry feature of the SocketIO upload client. This forced uploads to use only polling and prevented reconnections, removing a core part of the feature.
- **Subpath support in safe_mdv_app**: Added configuration so the Docker container can be run in a subpath by changing the setting `MDV_API_ROOT=/` in the docker YAML file.

Forcing polling-only causes issues in dev environments due to CORS restrictions on HTTP requests. WebSockets are not affected by CORS in the same way and already fall back to polling if WebSocket fails. Removing WebSocket impacts dev uploads unnecessarily.
The same issue is observed in the BIA environment, where WebSocket uploads don’t seem to work. I recommend checking the BIA server configuration and verifying the Vite proxy settings for polling. WebSockets provide a more reliable and performant connection. Removing them reduces the robustness of the feature, whose purpose was to guarantee reliable connections.

Testing: confirmed uploads work under these environments with files up to 15 GB:
- Docker default setup: frontend + backend on `localhost:5055`
- Dev frontend (`localhost:5170`) uploading to backend on `localhost:5055`
- Docker subpath setup: frontend + backend on `localhost:5055/server/`
- Dev frontend (`localhost:5170`) uploading to backend running with subpath config on `localhost:5055/test/`

PR #282, opened 2025-11-10
Md 676 points not appearing properly on viv chart

xinaesthete:
This mostly happened with images in SpatialData, where the physical scale metadata isn't found so it doesn't have a scale bar... the `scaleBarLayer` being `undefined` seemed to have some strange behaviour when included in the list of layers sent to deck.gl.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scale bar rendering to prevent undefined layer from affecting display.

* **Refactor**
  * Reorganized internal utilities and adjusted rendering configuration options.


xinaesthete:
I'm not sure this is 100% correct now - still some strange behaviour with deck sometimes, but it is better than it was so I'm merging.

PR #283, opened 2025-11-11
Fix violin plot min max not found issue

nawabfurquan:
## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric columns missing precomputed min/max no longer cause errors; min/max are computed and cached on demand.
* **Improvements**
  * Visualization scales and histogram bin ranges now use the on-demand min/max as fallback for color scales, box/violin plots, and range bins.
  * Runtime diagnostics emitted when min/max are calculated.

### DIFF
--- a/src/charts/HeatMap.js
+++ b/src/charts/HeatMap.js
@@ -17,7 +17,7 @@ class HeatMap extends SVGChart {
         });
 
         const p = this.config.param;
-        //! we used this.fieldNames instead of `param[1, ...]` 
+        //! we used this.fieldNames instead of `param[1, ...]`
         // to avoid mutating config - which is observable - in setFields - which may be in an action
         // also note that a lot of this was actually because of the way active links worked during development
         // this is less needed now (although we still have the reaction -> mutation issue)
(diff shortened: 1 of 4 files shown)


xinaesthete:
There are indeed quite a few other places where `col.minMax` is used directly. The hypothetical `@computed` property would take some care to make sure we were consistently using the same kind of column object representation and is not something I'm seriously considering right now (definitely over-engineering for this but there may be a revisit in future that it could fit in with).

I'm adding some changes, but it touches quite a few files and it would be nice to have test coverage for anything involved...

Also... I'm mostly considering `column.minMax` so far, but any use of `column.quantiles` is ultimately subject to the same set of issues, which may surface less frequently and be more likely to go unnoticed.

Places that need attention:
- `ChartManager._getColumnsFromOtherDataSource()` may be implicated of the root-cause of the bug we witnessed? This is only called from one place, but the things leading to that call are quite complex to follow.
- `HeatMap.js`
- `MultiBoxPlot.js`
- `DataStore.addColumn()` - this is where metadata may be added without the `minMax` being present - i.e. from a linked column where we don't have all of that pre-computed in `datasources.json`.
- `DataStore.getColumnRange()` - we should be able to use `getMinMaxForColumn()` here to avoid any issue. I had much earlier added a bad workaround to use a default [0, 50] range for `getColumnRange()` earlier.
- `DataStore.getColorLegend()` - there was also some typo `let range = c.minMaX` 
- `RangeDimension.getBins()` should be ok to call `parent.getMinMaxForColumn()` here.

PR #341, opened 2026-02-24
Fix unique value update issue

nawabfurquan:
## Summary by CodeRabbit

* **New Features**
  * Added support for a "unique" column type that stores/display strings; numeric inputs are coerced to strings and max string length is auto-managed.

* **Bug Fixes**
  * Improved validation and clear errors for missing/invalid string-length.
  * Safer decoding with bounds checks and null/empty handling to avoid corrupt or out-of-range displays.

* **Tests**
  * Extensive unit and end-to-end tests covering unique-column workflows and save-state behavior.

* **Documentation**
  * Added examples and CLI helpers to create projects with unique columns.


nawabfurquan:
> I've gone on a bit of a tangent here making some changes, trying to add tests and being clear under what circumstances the problem occurred. Haven't been able to figure out how to use the edit mode on the old `TableChart`, which seems like a code-path that would be relevant, need to try to wrap up what I have in my working copy and get this merged sooner rather than later.

I just added the `editable: true` to the unique column in the `datasources.json` to make it editable, but I believe I checked that in the new table not the old one to test it.

PR #362, opened 2026-03-13
Feat: View creation for new datasource

nawabfurquan:
## Summary by CodeRabbit

* **New Features**
  * Automatically create a combined view when adding datasources to projects that already have data.
  * File uploads now redirect to newly created views after completion; UI shows a “Redirecting to your new view…” state.
  * Processing flows now return and surface created view names so callers can navigate to new views.

* **Bug Fixes**
  * Added guards to avoid errors when datasource or view entries are missing during layout changes and chart rendering.


nawabfurquan:
@coderabbitai review

PR #366, opened 2026-03-18
UV migration

xinaesthete:
Main motivation to switch is for faster builds/install, as well as to introduce PR review workflow to Manjula (welcome to the team).

I've made it somewhat tied to the version of `uv` (0.9.25) that was already installed in the base `nikolaik/python-nodejs:python3.12-nodejs22` image. It won't have any major detrimental effects installing with a more recent version installed in the system, but running `npm run uv-setup` explicitly installs that version, and the CI action config etc also specify it.

Any users with existing environments outside docker will need to recreate their environments (`npm run python-setup` is still the canonical way to do this as far as I'm concerned). This is now using the `uv` default of `python/.venv`, so any `source path/to/mdv/venv/bin/activate` type workflow will need to be adapted to `source path/to/mdv/python/.venv/bin/activate` (I'll need to change some `~/.bashrc` on cluster accounts etc).

## Summary by CodeRabbit

* **Chores**
  * Switched the project from Poetry to uv for Python dependency management across CI, Docker, build scripts, and local workflows.
  * Updated build and run commands to use uv-based invocation and synchronization.
  * Adjusted package scripts and make targets to invoke uv tooling.

* **Documentation**
  * Revised developer and performance testing docs and READMEs to reflect uv setup and commands.

* **Chore / Config**
  * Migrated project metadata to a PEP 621-compliant configuration with dependency groups and a uv-compatible build backend.


xinaesthete:
Sadly, I have to report that having rebuilt the container without cache a couple of times, the `uv` version actually seems to be taking ~33% longer than `poetry`. Given that this was the main concrete hypothetical benefit, I'm not sure what to make of it.

It seems vaguely nice that it uses `pyproject.toml` with a PEP-standard rather than Poetry which is a bit more bespoke... but that's not a strong reason to change.

PR #424, opened 2026-04-22
codex/discard unsaved added columns on view switch

nawabfurquan:


nawabfurquan:
Should be merged after #423

PR #435, opened 2026-04-30
PostgreSQL replacement with SQLite 

lferreira86:
Branch to test replacement of PostgreSQL with SQLite backend database. 
* Update deploy scripts (.bat .sh and _gui.sh) to create dockerfile and env file to facilitate deployment
* Changes on database ORM to replace PostgreSQL with SQLite 

## Summary by CodeRabbit

* **New Features**
  * Interactive local Docker deployment across `deploy.sh`, `deploy.bat`, and `deploy_gui.sh`, with selectable DB backend (sqlite/postgres), new vs replace/reuse modes, and automatic host port selection
  * Runtime support for `DB_BACKEND` with improved PostgreSQL readiness and optional schema setup
  * New CLI to export user–project permission data to stdout or CSV, with optional Auth0 login enrichment
* **Documentation**
  * Added a “Deployment Scripts (Local Docker)” guide with parameters and multi-deployment examples
* **Tests**
  * Expanded unit test coverage for sqlite/postgres configuration and startup behavior
* **Chores**
  * Ignore deployment-generated override/env artifacts; added Docker Compose orchestration files


MarShaikh:
Hi @lferreira86, I was fixing some small issues raised by the CI and ran into these while trying to run the `deploy.sh` script locally. 
1. The `if [[ "$ports_line" =~ (0\.0\.0\.0|\[::\]):([0-9]+)->5055 ]]` in `deploy.sh` gave me a syntax error:
`conditional expression: unexpected token '>'`. So I fixed it by moving the pattern into a variable, 
```bash
local port_pattern='(0\.0\.0\.0|\[::\]|:::):([0-9]+)->5055' 
if [[ "$ports_line" =~ $port_pattern ]]
```
2. Deploy and build use different compose files. `deploy.sh` runs against `docker-local.yml`, while `docker compose build` / `make` use `docker-compose.yml`. Is that split intentional, or could the deploy use the same compose file as the build?
3. `docker-compose.yml` and `docker-local.yml` currently differ only by port (`5056` vs `5055`). Is there a reason to keep both, or could they be merged into one?
4. The compose files load `.env`, while `deploy.sh` writes `.env.<name>`. I wanted to understand this better, is this a shared `.env` for secrets plus a per-deployment `.env.<name>` for config, or should the compose files reference `.env.<name>` directly?

lferreira86:
> Hi @lferreira86, I was fixing some small issues raised by the CI and ran into these while trying to run the `deploy.sh` script locally.
> 
> 1. The `if [[ "$ports_line" =~ (0\.0\.0\.0|\[::\]):([0-9]+)->5055 ]]` in `deploy.sh` gave me a syntax error:
>    `conditional expression: unexpected token '>'`. So I fixed it by moving the pattern into a variable,
> 
> ```shell
> local port_pattern='(0\.0\.0\.0|\[::\]|:::):([0-9]+)->5055' 
> if [[ "$ports_line" =~ $port_pattern ]]
> ```
> 
> 2. Deploy and build use different compose files. `deploy.sh` runs against `docker-local.yml`, while `docker compose build` / `make` use `docker-compose.yml`. Is that split intentional, or could the deploy use the same compose file as the build?
> 3. `docker-compose.yml` and `docker-local.yml` currently differ only by port (`5056` vs `5055`). Is there a reason to keep both, or could they be merged into one?
> 4. The compose files load `.env`, while `deploy.sh` writes `.env.<name>`. I wanted to understand this better, is this a shared `.env` for secrets plus a per-deployment `.env.<name>` for config, or should the compose files reference `.env.<name>` directly?

@MarShaikh, please check my answers below:
1 - Ok, i think this is a better solution;
2 and 3 - The deploy.sh prepare the deployment file, i don't remember which name it use for the final composer file, if docker-local or docker-compose, however we should merge this so there is no confusion with the deployment. The port is controlled as well by the deply.sh, it get the next available port and configure the deployment;
4 - the .env file is per deployment, because we can use different configuration for SQLite/PostgreSQL, including database reuse, shcema names, ports...

PR #459, opened 2026-05-15
Option to show scatterplot density layers as grid rather than layers on same plot

xinaesthete:
Intention in future is to have more general ways of showing several plots in a single chart (with lazy-loading).

For now, this is a first-version which just adds a new mode in "scatterplot with density".

## Summary by CodeRabbit

* **New Features**
  * Density-grid visualization for scatter charts with a user toggle for "overlay" vs "grid" modes.
  * Two new density-grid viewers (Deck + Viv) rendering per-cell density overlays, labels, and synchronized per-cell view state.
  * Responsive chart-array layout and hooks for measuring, visibility, geometry, and stable multi-viewport routing.

* **Bug Fixes**
  * Fixed a CSS termination issue affecting chart container styles.

* **Tests**
  * Added unit and end-to-end tests covering density-grid behavior, layout utilities, visibility, and selection UI.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Taylor-CCB-Group/MDV/pull/459)


xinaesthete:
I think there is currently a bug with the gate layers which means that it is appearing more intense, probably because it's making a clone for each view and then not filtering the ones from outside the view (or something). You can see this change as you scroll and a different number of views are active...

PR #460, opened 2026-05-18
Background filters

martinSergeant:
### Purpose
The purpose of this branch was improve background filters so thry could display bucket/cell type specific umaps and dotplots. In doing this bugs were found in the dotplot, which was corrected

#### Background Filters
* Background Filters apply at the chart level and are permanent filters i.e. are independent of the charts's local filter and global filters.

* They were used initially only in scatter plots to show specific regions in spatial data, but there was no way for a user to add/modify them.

* Before, a dimension would have a separate array to store the background filter but now it is integrated into the local filter. This way no extra data has to be passed to the worker. Local filters can how have fourth state (local filter + background filter)

    * 0 - no filter
    * 1 - local filter
    * 2 - background filter
    * 3 - local filter + background filter

They are applied with setBackgroundFilter and clearBackgroundFilter of Dimension. At the moment, only one column can be used and it has to be text ,text16 or multitext. 

The filterPredicate function of the base dimension now takes background filters into consideration.

Individual charts/dimensions have to decide how they are going to handle the background filter when aggregating/displaying data. To date only dotplot and the classic scatterplot implement background filters.

#### Dotplot

* A bug was fixed in the dotplot so the correct value was displayed (before the average was taken from only those items with data , now all the items in the group are used to calculate the average)
* An option of a different color scheme has been added
* The size legend has been improved so the legend items do not overlap 

## Summary by CodeRabbit

* **New Features**
  * Background filters that persist across charts and merge with local filters
  * DotPlot: row/column clustering with dendrograms, selectable color palettes, and Y-axis ordering
  * Settings: background filter controls and improved color palette options

* **Improvements**
  * Color legend spacing, sizing and tick customization; fraction legend better normalized
  * Highlights respect current local filtering
  * Build can produce versioned output filenames
  * Project viewer accepts HTTP URLs as project dirs
  * DotPlot y-axis category type widened; removed Averaging Method control

* **Documentation**
  * Added Background Filters documentation

* **Tests**
  * Added background-filter unit tests and a routing test

[Review Change Stack](https://app.coderabbit.ai/change-stack/Taylor-CCB-Group/MDV/pull/460?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/datastore/Dimension.js
```diff
@@ -236,31 +239,64 @@ class Dimension {
...
+        if (!this.bgfData) {
+            return;
+        }
         this.bgfData = null;
+        for (let i = 0; i < this.parent.size; i++) {
+            if (this.filterArray[i] > 1) {
+                this._applyStateTransition(i, this.filterArray[i] - 2);
+            }
+        }
     }
```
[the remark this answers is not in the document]
martinSergeant: This was fixed by renaming all in instances to clearBackgroundFilter, which is in step with setBackgroundFilter

src/datastore/Dimension.js
```diff
@@ -37,21 +58,26 @@ class Dimension {
...
-                        parent.filterSize++;
-                    }
+            //not interested in rows already in background filter
+          
+            const shouldExclude = !predicate(i);
+            const prevState = localFilter[i];
+            // When a local filter is active, background-hidden rows must also
+            // contribute to exclusion scope (promote 2 -> 3 once).
+            if (prevState === 2) {
+                this._applyStateTransition(i, 3);
```
xinaesthete, line 66: I'm not super-keen on these kinds of side-effects here, thinking about the intention to be able to make these things async etc...

I suppose we can come back to this later.

Could we not also exit early for rows already in background filter, without needing to call `predicate`?

xinaesthete: It looks as though we could just move the cases that don't depend on `shouldExclude` to before we do `const shouldExclude = !predicate(i)` and it should be able to save significant amounts of work processing rows in the background filter.

That looks safe without any risk of side-effects in `this._applyStateTransition` being relevant.

As long as the `predicate` itself is side-effect free, of course.

src/charts/WGLScatterPlot.js
```diff
@@ -693,6 +737,41 @@ class WGLScatterPlot extends WGLChart {
                 },
             });
         }
+        const backgroundColumnSetting = {
+            type: "dropdown",
```
xinaesthete, line 741: We could use `type: "column"` here; although it does seem we need to consider more generally adding a `"None"` option.

xinaesthete:
I tried using a `DotPlot` with a background-filter on the `gates` column and it didn't seem to work. Could be a general issue with `'multitext'`?

edit: it worked with the kind of `'multitext'` saved by my 'tag annotation' dialog (that I've utterly failed to get anyone to attempt to use or give feedback on, probably does need to be made simpler somehow to be fair), but not gates (and I'd suspect not others which use the other kind of encoding). Sidenote: I must remember to review previous changes I made relating to this but never merged.

xinaesthete:
I had been thinking along the lines of having something that would allow to have essentially a copy of `SelectionDialog`, perhaps as another tab in `SettingsDialog`.

We could possibly change how we serialise `background_filter` so that it's in a form more similar to that (even if the way we expose the UI) but I suppose we can always have a step for migrating configs from this version to that when we get around to it.

src/datastore/Dimension.js
xinaesthete: The side-effects can stay, but I'll be hoping to make some considerable changes to how `Dimension` related functionality works at some point so that

- `async` operations (i.e. workers and/or WebGPU) are able to work
- filters can be composed in different ways so that as well as for the whole DS or a particular chart (different layers within a chart, charts sharing some filter state, etc)

So that'll be a potentially different thing and we can consider how it gets implemented, assuming it will involve a bit of upheaval. At least more state change is managed within `Dimension.filterPredicate` here so that should mean that we have something to reason with when it comes to doing something different.

I do think we should try to fix the bug with `gates` though as this seems like something people would reasonably want to use with this feature.

martinSergeant: When I use gates for a background filter I don't experience any problems-  what exactly is the issue/error ?

 _we could just move the cases that don't depend on shouldExclude to before we do const shouldExclude = !predicate(i)_
This is true, but there was a bug in the current code which manifests when there is a local  filter on a chart with a background filter and that background filter is removed, The background filtered rows are still contributing to the local  filter due to local membership for background-hidden rows becoming stale. Hence I needed to change filterPedicate and now predicate has to be run on background filtered rows.

There may be a simpler way of doing it, but a complication is that if background filtered rows are included in the fiiter e.g. within the polygon in a scatter plot. they still need to be filtered out  as it gets confusing because you would be including rows (points) you can't see on the chart

martinSergeant: Actually I have done as you suggested, the bug was in clear/set background filter not taking into consideration local filters. This has been corrected and the cases that don't depend on should exclude moved to before predicate is called

Also filters (normal and background)  can be a bit confusing with multitext - if  you select a categories for the dotplot then other categories could also be shown because if they co-occur in a row with selected categories. In the genome branch I have heatmap which works with multitext and this filters out categories at the chart level , which fixes this

Was this the problem with  gates  ?

PR #465, opened 2026-05-18
Fix chart loading issues by adding early checks and better error messages

nawabfurquan:
## Summary by CodeRabbit

* **New Features**
  * Debug UI “Fault Injection” to simulate empty, corrupted, or timed-out data responses.
  * One-shot front-end fault injection applied to the next data fetch via session storage.

* **Bug Fixes**
  * Faster detection and retry of missing chart data fields with clearer, richer error reporting.
  * More reliable tracking of initial chart/view-loaded state during startup and failures.

* **Tests**
  * E2E test updated to stub project resources and wait for chart readiness.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Taylor-CCB-Group/MDV/pull/465?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/charts/ChartManager.js
+++ b/src/charts/ChartManager.js
@@ -1445,6 +1445,7 @@ export class ChartManager {
         const lc = this.config.dataloading || {};
         split = lc.split || 10;
         threads = lc.threads || 2;
+        const retries = lc.retries ?? lc.maxRetries ?? 2;
         this.transactions[id] = {
             callback: callback,
             columns: [],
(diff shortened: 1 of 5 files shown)


xinaesthete (commented):
Potentially we could have tried to avoid adding more `ChartManager` data-loading lifecycle stuff, we may in future still want to refactor a bit and there's a chance that not all column-loading goes via the route we added here... so those might not be able to benefit in the same way, but I think this is basically fine and not worth getting hung up on here.

PR #470, opened 2026-05-21
Genome viewer

martinSergeant:
This branch adds the following:- 
* A new 'genome'  parameter in datastore which means the datastore represents genomic location information
* SVCircosPlot-  a circos plot for displaying structural variants genome wide
* UCSC browser - displays an image o af UCSC genome  browser session at the supplied genome position
* IGVBrowser - embedded igv.js component whose default track can display the datastore information directly, no track file derived from the datastore is required
* CategoryHeatMap - A heatmap for displaying two categorical columns (text/multitext or text16) 

## Summary by CodeRabbit

* **New Features**
  * Added IGV and UCSC genome browser charts for genomic data visualization
  * Added SVCircosPlot chart for structural variant visualization
  * Added CategoryHeatmap chart for visualizing co-occurrence patterns
  * Added Python tools to import structural variant VCF files and map genomic features
  * Extended server to proxy UCSC Genome Browser images and serve external track files

* **Documentation**
  * Added genome metadata specification for interval and structural variant data


src/charts/SVCircosPlot/SVCircosPlot.js
```diff
@@ -0,0 +1,72 @@
...
+    init:(config,dataSource)=>{
+        const cols= dataSource.genome?.svs?.sv_columns;
+        if (!cols){
+            throw Error("Trying to add SVCircosPlot, but No SV columns found in datasource genome metadata")
+        }
+        const chromosomes = dataSource.genome?.chromosomes;
+        if (!chromosomes || Object.keys(chromosomes).length === 0){
+            throw Error("Trying to add SVCircosPlot, but No chromosome metadata found in datasource genome")
+        }
+        config.param=[cols["pos1"],cols["chr1"],cols["pos2"],cols["chr2"],cols["svtype"]];
```
xinaesthete: Hypothetically here we could just do

```ts
config.pos1 = cols["pos1"]

...

const pos1 = useFieldSpec(config.pos1);
```
etc

But actually, it seems to me that this stuff isn't really part of the 'config' at the chart level, it's something that's derived from the datasource metadata and as such it may as well just be local state that gets pulled from that inside the component itself?

So I'd suggest something like

```ts
const { pos1, chr1, pos2, chr2, svtype } = useStructuralVariance();
```

and aiming for things that live on the `config` object being new information needed for the individual chart.

Relates to my canonical example of viv charts, where the only bit of information expressed in the "Add Chart" dialog is which region to use, which then gets turned into a fairly complex config object describing all the `param`, `background_filters` etc. This then means that if we want to write those config objects from Python, the implementation there needs to know how to synthesise the appropriate information and is liable to get out-of-step if the frontend changes.

The `useStructuralVariance()` implies that there might be some `SVContextProvider`, or just

```ts
function useStructuralVariance() {
  const ds = useDataSource();
  if (!ds.genome) throw new Error("trying to use structural variance on incompatible datasource);  
  // actually I've realised this is wrong, I was thinking `sv_columns` was an array...
  // and then I was going to suggest it be more of a structured object
  // const [pos1, chr1, pos2, chr2, svtype] = useFieldSpecs(ds.genome.svs.sv_columns);
  // return { pos1, chr1, pos2, chr2, svtype };
  // which I now realise it already is... so I'd instead suggest that we have real column objects on it
  // make sure they're loaded (and any other necessary guards) then just
  return ds.genome.svs.sv_columns;
}
```

We should also have datasource schema for `genome` (maybe that was already there I can't remember).

<sub>Commented from [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/Taylor-CCB-Group/MDV/pull/470)</sub>

xinaesthete: Another benefit of not keeping more information than we need on the `config` object is that we're more likely to be able to do things like copy a chart from one project to another... at least, it reduces the surface-area of what needs to be checked during an operation like that.

xinaesthete: p.s. we have 'copy chart config' labeled as a debug feature in the hamburger menu, but actually it might be potentially viable to be able to copy & paste (even between projects with different datasource configurations if the configurations are sufficiently well-matched) if we're a bit mindful of this stuff.

xinaesthete:
Meant to comment earlier; one remaining niggle is the new methods in `serverlite.py` which I'm worried about meaning things wouldn't work properly elsewhere.

There's a `class MDVProjectServerExtension` in `server_extension.py` which we can use for declaring endpoints relating to a particular feature, this is precisely the kind of thing it's intended for.

src/charts/GenomeBrowsers/SVCircosPlot/SVLayer.js
```diff
@@ -0,0 +1,247 @@
...
+import { Model, Geometry } from "@luma.gl/engine";
+
+/* I could use deck.gl's filtering extension but it 
+ it appears to work in a similar way shader-modules.ts line 241
+    if (dataFilter_value == 0.0) {
+       gl_Position = vec4(0.);
+    }
+ at the end of the vertex shadwer 
+ So if we do it at the beginning and return early
+ it should be more efficient
```
xinaesthete, line 12: this might be something we should mention to deck.gl community

xinaesthete:
I'm still concerned about the `/ucsc_proxy` handling. I've made some changes which I'll push that bring that route into somewhere that our other server implementation can reference, but we also need to consider that we often aren't mounted on the root of a domain so we need to use `MDV_API_ROOT` based URLs, not `/*`. I have changes which I'll push that should be right, but since I don't really have a way of verifying that I'm reticent to merge without @martinSergeant checking it looks ok.

I also noticed biome is flagging some `useExhaustiveDependencies` issues which in future we might try to get more strict on (they can flag genuine issues, and if we have lots of warning noise as we probably generally do then it's hard to notice).

martinSergeant:
All seems to be working fine so you can merge it -
I have attached a tiny test project
[test_genome_project.zip](https://github.com/user-attachments/files/29841025/test_genome_project.zip)

xinaesthete:
> All seems to be working fine so you can merge it - I have attached a tiny test project [test_genome_project.zip](https://github.com/user-attachments/files/29841025/test_genome_project.zip)

Great, I hadn't noticed your reply and I was just going to comment that something like that would be useful. I trust that if you say it's working then it should be fine, I'm eager to try the test project but I'm going to merge and then have a look from there to save a bit of faff; we can always patch any issues later.

PR #477, opened 2026-06-01
Consolidate mdvtools packaging: one pyproject, slim core + single app extra

MarShaikh:
## Summary

  Collapses `mdvtools` to a **single `pyproject.toml`** (on uv, building on #472) so we stop
  maintaining two diverging build configs. A bare `pip install mdvtools` is now the **slim
  core** (project manipulation, viewing, `MDVProject.serve()`, and spatial); the heavy
  server / chat / auth stack moves to one optional **`app` extra**, guarded at import. The
  redundant `mdvtools_lite_build/` (Hatchling + `force-include`) is deleted.

  `pip install mdvtools`          → `slim core`           (the majority of users)
  `pip install "mdvtools[app]"`   → `full app`            (database, chat/LLM, auth)
  `uv sync --group dev --extra app`  → what `Docker`/`CI` build

  Why this shape (full rationale in `docs/adr/0001`, `docs/adr/0002`, and
  `docs/MDVTOOLS_PACKAGING_PROPOSAL.md`):
  - We ship all modules to `mdvtools` "slim"  and gate with guarded imports, matching popular frameworks like
    pandas/transformers/scanpy.
  - `auth`/`dbutils`/`llm` are a runtime import cycle → modelled as **one** `app` extra, not
    three.
  ## What changed

  - **`pyproject.toml`**: move `chat/server/auth` deps out of core; fold the `backend`/`llm`/`auth`
    PEP 735 groups into a single `[project.optional-dependencies].app` (+ `all` alias); keep
    `dev`/`docs` groups. Classification fixes (verified against runtime imports):
    - `ruff` → `app` **and** `dev` — it's a *runtime* dep (chat shells out to `ruff check` in
      `llm/code_manipulation.py`), and also used by `make lint`.
    - `generate-tiff-offsets` → core (`spatial/jp2k.py` imports it; spatial is core).
    - `matplotlib`/`regex`/`dotenv`/`nbformat`/`tabulate`/`gspread`/`oauth2client` +
      `langchain-community` → `app`.
  - **Guards**: new `mdvtools/_optional.py` + guards on 6 leaf modules
    (`llm/{langchain_mdv,chatlog,github_utils}`, `dbutils/{dbmodels,mdv_server_app}`,
    `auth/auth0_provider`) → friendly `pip install "mdvtools[app]"` error on a slim install.
    Covered by `tests/test_optional.py` (helper + a static contract that each heavy leaf stays
    guarded).
  - **`uv.lock`** regenerated
  - **Dockerfile + CI** (`py_ci` ×2, `py_performance`): `uv sync` → `--group dev --extra app`
    (removed `backend`/`auth` group references).
  - **Deleted `mdvtools_lite_build/`**; salvaged its content into `python/PYPI_README.md`
    (now the PyPI long description, `readme = "PYPI_README.md"`) and `docs/RELEASING.md`
    (uv `build`/`publish` workflow).
  - **Docs**: `CONTEXT.md` glossary, ADR-0001, ADR-0002, packaging proposal.

  ## Behavior changes / notes

  - **Breaking for bare-install users of chat/server/auth**: those now require
    `pip install "mdvtools[app]"`. Spatial is unchanged.
  - The **PyPI page renders `PYPI_README.md`**, not the dev-facing `README.md`.

  ## Testing

  - Full env: `make test` → **186 passed** (incl. the guard suite).
  - Slim clean-room: `uv build` (451 static files bundled) → `pip install` the wheel into a
    fresh venv outside the source tree → quickstart ran → `import mdvtools.llm.langchain_mdv`
    raised the friendly `mdvtools[app]` error.

  ## Not in this PR (follow-ups)

  - [ ]  Publish to TestPyPI → PyPI per `docs/RELEASING.md`.
  - [ ] User-facing release note / version bump.

## Summary by CodeRabbit

* **New Features**
  * Optional `app` extra available via `pip install mdvtools[app]` for full feature set including database, authentication, and LLM functionality

* **Documentation**
  * Added PyPI README with installation instructions and quickstart examples
  * New packaging and release guidelines

* **Chores**
  * Migrated to unified `uv`-based packaging and build system
  * Reorganized dependencies with slim default installation and optional extras for advanced features


docs/MDVTOOLS_PACKAGING_PROPOSAL.md
```diff
@@ -0,0 +1,190 @@
...
+| What | Size | Notes |
+|---|---|---|
+| `dbutils` + `auth` + `llm` **.py source** (what the lite build actually excludes) | **~576 KB** | the modules in question |
+| `mdvtools/static` (built JS frontend) | **59 MB** | already shipped in the slim wheel today, in both builds |
+| Full install footprint | **~1.2–1.4 GB** | matches Martin's number |
+| └ dominated by deps: faiss, langchain*, scanpy, spatialdata, … | **≈ all of it** | gated by the `app` extra |
+
+**Correction to an earlier framing:** the lite build does **not** exclude `spatial`. Its `force-include` ships `mdvtools/spatial` and it hard-requires `spatialdata`/`spatialdata-io`/`ome-zarr`. So today's PyPI wheel already includes spatial. What the lite build excludes is `dbutils`/`auth`/`llm`. Excluding those saves ~576 KB of ~1.2 GB ≈ **0.05%** — rounding error. (See ADR-0001.)
+
+Every comparable project ships all its modules and guards imports — pandas ships `io.excel` without openpyxl; transformers ships every model without torch; scanpy ships everything. None maintain a second toml to trim KB.
```
xinaesthete, line 49: @martinSergeant Can you check this PR generally seems reasonable, are your concerns met?

xinaesthete (approved):
Looks good to me.

docs/MDVTOOLS_PACKAGING_PROPOSAL.md
martinSergeant: It looks good

Pulling in only certain local  modules does not save that much space, so a single toml file/build is the best way forward

I built and installed both versions. The only  problem I had was when  pip installing the app version - pyscopg would not compile - I had to do the following first
```
sudo apt install libpq-dev python3-dev
```
Not sure whether having psyocpg2-binary as dependency instead  would work or whether we are even going to need it at all now that SQLite is being used.

RELEASNG.md seems pretty comprehensive - tagging with the version number in the toml file is probably the best solution - I was trying to come up with an automated way of doing this but none were practical

The app version was only 300 MB larger than standard one , so we're not saving as much. as I thought.  Other ideas for the future

* not having maps and excluding the examples dir in the  in the JS build  (saves about 40 MB)
```
nomap=true exclude_dir=true pnmp run build-flask-vite
```
* do we need polars and pandas ?
* conditional  inclusion of spatialdata dependencies (saves 600 MB)

xinaesthete: I think we could try to have some kind of somewhat informal protocol around any import we don't guarantee to be there, such that we can easily find and reason about these places in the code and return clean errors to users.

I think we can merge this, just wanted to make sure you were aware and happy with it @martinSergeant

xinaesthete: As for source-maps; we have various error-widgets etc that I suppose would be somewhat degraded without, but we could still be able to reconstruct pretty stack-traces without (as in, hypothetically we know what build they're based on and if we needed to understand an error we could theoretically make a build with map from the given commit but this all sounds like a faff).

PR #478, opened 2026-06-02
Hard compound columns

martinSergeant:
Allows the user to create 'compound columns' (merge columns with a separator) using the add column workflow of  the table chart

The add column dialog enables merging of columns as well as cloning a single column to produce the new column

This creates a 'hard' column' which will not update if any of the source columns change

This is separate to the intended feature of adding virtual/computed columns

## Summary by CodeRabbit

* **New Features**
  * Added "compound" column creation to combine multiple text/text16 columns with a configurable delimiter and mode selector in the add-column dialog.
  * Dialog now presents mode selection and conditional inputs for compound setup; submission validates required fields.

* **Bug Fixes**
  * Enforced at least two text source columns for compound creation; auto-upgrades to text16 when distinct-value limits require it; prevents creation when distinct-limit exceeded.

* **Tests**
  * Added unit tests covering compound creation, loading of unloaded sources, validation failures, and distinct-value behaviors.


martinSergeant:
@coderabbit review

xinaesthete:
I don't mind too much either way here, but for future 'computed' columns, I would have thought that the newly generated columns would *not* be editable, in the sense that they will be strictly a function of whatever they're computed from (so then if they get manually edited there's not a clear source-of-truth).

So I'd quite like a version of this where the result *isn't* editable and we add some new metadata to the column in `datasources.json` (edit: can't seem to get github to format this properly)

```json
{
  '...': 'existingColumnMetadata',
  'provenance': {
     'dynamic_op': 'compound',
     'separator': '_',
     'source_columns': [
        'disease_state', 'cell_type'
     ]
  }
}
```

and the extra metadata could basically be ignored in the short-term, but we can have a schema for it and then it can serve as a prototype for the derived version in a way that we can experiment more with the UI and have a functionally equivalent way of having synthesised on-demand or kept in the store etc.

If we expect that we *do* want to edit the generated column, in that case we explicitly chose that as an option and make sure that it doesn't pretend to be something that could be derived from some `source_columns`.

Having said that, it all gets more complicated as soon as any `source_column` is edited or there's a loop in the `source_columns` dependencies. But all of this is for more general consideration later/elsewhere.

Anyway, I'm not averse to including the version we have here, I'd be tempted to start adding somewhat ad-hoc metadata (along with some schema) but there's probably something to be said for that being done later.

xinaesthete:
I think I'll go ahead and merge this so people can use it and we can follow up with other things later.

PR #487, opened 2026-06-04
Add bulk delete, recycle bin and restore project

nawabfurquan:
## Summary by CodeRabbit

* **New Features**
  * Move projects to a recycle bin, restore them, and permanently purge deleted projects
  * Bulk select projects and perform batch "Move to Recycle Bin" actions
* **UI Changes**
  * Recycle Bin dialog with list, restore and "Delete all permanently" flows; project cards/list support selection checkboxes and updated delete wording
* **Tests**
  * Added unit and end-to-end tests covering recycle-bin, bulk delete, purge and restore behaviors

### DIFF
--- a/python/mdvtools/dbutils/cleanup_projects.py
+++ b/python/mdvtools/dbutils/cleanup_projects.py
@@ -24,6 +24,7 @@ from typing import List, Dict, Tuple, Optional, Any, Union
 from datetime import datetime
 
 from mdvtools.dbutils.dbmodels import db, Project, File, UserProject
+from mdvtools.dbutils.dbservice import ProjectService
 from mdvtools.logging_config import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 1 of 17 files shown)


xinaesthete:
This seems good, I'm going to suggest @MarShaikh also has a look because I know there was a potential change to `dbmodels.py` definitions that would likely mean this implementation might be slightly different. Actually, maybe the code that used it would still be the same, just with better typing, not sure.

I'm basically very happy to merge this though, and I don't think it'll be difficult to adapt anything that may be necessary.

MarShaikh:
> This seems good, I'm going to suggest @MarShaikh also has a look because I know there was a potential change to `dbmodels.py` definitions that would likely mean this implementation might be slightly different. Actually, maybe the code that used it would still be the same, just with better typing, not sure.
> 
> I'm basically very happy to merge this though, and I don't think it'll be difficult to adapt anything that may be necessary.

A bit late to this conversation, but yes, the issue change was only related to type hints and not any logical change to the code.

PR #488, opened 2026-06-04
Release 1.3.0

MarShaikh:
Two changes in this PR:
- `1.2.3` -> `1.3.0` version bump in the `pyproject.toml` 
- `wheel-exclude/source-exclude` fix to reduce the size of the wheel
- A new `CHANGELOG.md` file

## Summary by CodeRabbit

* **New Features**
  * SlickGrid-based editable data table, column management (including compound columns)
  * New charts: density chart and Splatter Plot
  * Gating/permissions UX, GUI permission icon, gate drawing and management dialog
  * ChatMDV verification/evaluation tooling and opt-in app install option
  * Multiple conversion/upload improvements

* **Notable Changes**
  * Frontend toolchain upgraded; packaging and Python support updated
  * Updated home navigation routing and chart loading behavior

* **Bug Fixes**
  * Fixed chart rendering/interaction and selection issues
  * Fixed routing, TLS, preview/modal visibility, and table/datasource behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,12 +13,12 @@ All notable changes to `mdvtools` are documented here. The format is based on
 > stores; pass the new `--batch` flag to restore that directory-scanning behaviour. Update
 > any scripts or commands that pointed at a parent folder (#412).
 
-A feature release: a new react data table, gating, new chart types, ChatMDV evaluation
+A feature release: a new React data table, gating, new chart types, ChatMDV evaluation
 tooling, and a large batch of fixes and dependency upgrades. Packaging was consolidated
 internally — the published `pip install mdvtools` is the same slim/lite core as before.
 
 ### Added
-- New **SlickGrid-based react data table**: inline cell editing, find-and-replace,
+- New **SlickGrid-based React data table**: inline cell editing, find-and-replace,
   multi-row highlighting, and sortable columns (#328).
 - **Gating** — draw rectangle/polygon gates on scatter and spatial (viv) plots to define,
   name, colour, and manage cell populations (gates become a filterable column), with a
@@ -33,7 +33,7 @@ internally — the published `pip install mdvtools` is the same slim/lite core a
   merge spatial (Xenium) annotations, patching `cell_id` (#444); a SpatialData conversion
   report runner (#412); tsv/tab file upload with file-extension hints (#405, #408).
 - App/UX: Settings dialog with a General folder and search (#382); automatic view creation
-  when a datasource is uploaded (#362); Enter-key submission in view dialogs (#377); react
+  when a datasource is uploaded (#362); Enter-key submission in view dialogs (#377); React
   Text Box with collapsible markdown and mermaid diagrams (#425); subgroup selection in the
   link UI (#457); background filters (#460); schema-validation logging UI (#384); and
   category-selection widgets plus improved multitext/tag-annotation handling (#399, #401).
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,61 @@
...
+## [1.3.0] - 2026-06-04
+
+A feature release: new chart types, table editing, gating/permissions, ChatMDV
+evaluation tooling, and a large batch of fixes and dependency upgrades. Packaging was
+consolidated internally — the published `pip install mdvtools` is the same slim/lite
+core as before.
+
+### Added
+- Gating feature and access controls, plus permission display in the GUI (#328, #388, #389, #359).
+- New chart types: DeckSplatter (#391) and deck scatter density with shared contour
```
xinaesthete: `DeckSplatter` is a bit of an odd/experimental one, probably needs a bit of work to be useful... but also from a user perspective, it might be better to refer to the names as they appear in the UI, like "Splatter Plot" in this case.

CHANGELOG.md
```diff
@@ -0,0 +1,61 @@
...
+  settings and log bandwidth control (#390).
+- Table editing in the new react table: add, bulk-edit, remove, and rename columns, plus
+  hard compound columns (#386, #436, #478).
+- Conversions: X-based UMAP & Leiden options for AnnData/SpatialData (#392), Xenium
```
xinaesthete: Relevant to mention here is that semantics of `mdvtools convert-spatial` are significantly different in this version. Rather than requiring the user to point to a parent folder which contains some spatialdata zarr stores, it defaults to accepting path to the store itself, with a new batch option that's more like the previous (confusing and error-prone) behaviour.

CHANGELOG.md
```diff
@@ -0,0 +1,61 @@
...
+evaluation tooling, and a large batch of fixes and dependency upgrades. Packaging was
+consolidated internally — the published `pip install mdvtools` is the same slim/lite
+core as before.
+
+### Added
+- Gating feature and access controls, plus permission display in the GUI (#328, #388, #389, #359).
+- New chart types: DeckSplatter (#391) and deck scatter density with shared contour
+  settings and log bandwidth control (#390).
+- Table editing in the new react table: add, bulk-edit, remove, and rename columns, plus
+  hard compound columns (#386, #436, #478).
```
xinaesthete: Maybe rather than "hard compound columns" it's better to describe this differently here.

CHANGELOG.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Changed
+- Consolidated the previously separate `pyproject.toml` files (the lite build under
+  `mdvtools_lite_build/` and the full/dev project) into a single `python/pyproject.toml`.
+  `pip install mdvtools` continues to install the same slim/lite core; the heavy
+  server/chat/auth stack now lives behind the `app` extra, guarded at import (#477).
+- Build and publish migrated from Poetry/Hatchling + twine to **uv** (#472).
+- Supported Python is now `>=3.11,<3.13`.
+- Frontend toolchain: upgraded to Vite 8 (#442), migrated to pnpm (#441), enabled the
+  React compiler (#443), and moved CI to Node 24 (#455); various dependency upgrades
```
xinaesthete: React compiler is still disabled by default, but also I'm not sure if all of this is of interest to consumers of the python package.

MarShaikh: Yes, definitely. I thought since none of this is going to be visible on PyPI, the actual python package consumers, maybe this could be there in the changelog for the repo?

xinaesthete: I misunderstood; that's fine, but like I say – technically not true to say react compiler is enabled. We have the means to enable it (and actually, maybe I will do that – no known issues with it and I think it should be ok)

xinaesthete (approved):
LGTM, thanks.

PR #490, opened 2026-06-04
Viv UBO update

xinaesthete:
Updating to the new version of viv with associated changes to shader-extensions. Now support up to 10 channels (arbitrary limit imposed in viv, might change in future). The shaders also no longer do all of the work associated with channels they're not using.

Before merging, I should do a pass on "Image Array" to see whether it's still working. This requires a somewhat obscure project datasource configuration (thankfully I was doing some housekeeping and stumbled on such an example recently).

## Summary by CodeRabbit

* **New Features**
  - Support for more color channels and per-channel brightness/contrast controls with improved reset behavior

* **Bug Fixes**
  - More robust spatial selection and editable-layer handling (safer feature acceptance and type guards)
  - Improved translation and rectangle-drawing interactions

* **Rendering Improvements**
  - Improved heatmap/contour and saturation shader wiring; texture array sizing/depth handling enhanced

* **Chores**
  - Internal compatibility and typing cleanups for editable layers and state handling


xinaesthete:
Before merging we need to check for any regressions in mouse-event handling, particularly with transitions between fullscreen/popout.

PR #493, opened 2026-06-08
codex/add filtering to continuous legend

nawabfurquan:
## Summary by CodeRabbit

* **New Features**
  * Color legends now support continuous numeric-range filtering with interactive drag selection, clear/commit behavior, and a dedicated drag handle.
  * Continuous legends accept controlled range props so external UI can reflect or control the active selection.
  * Improved continuous-legend sizing and label layout for more reliable rendering.

* **Refactor**
  * Shared brush/scale utilities introduced to unify histogram brushing and interaction.

* **Tests**
  * Added unit, integration, and end-to-end tests covering continuous-range interactions, persistence, clearing, and reset UI.


xinaesthete:
There is a `useBrushX` hook that is currently declared in `HistogramWidget.tsx` but which might be good to re-use here. I prefer to be able to e.g. drag to sweep a range of a particular width, and the UX we have here is inconsistent with other places where there are similar interfaces (including the "Histogram" chart which currently has a separate vanilla d3 implementation).

When I say "might be good to re-use here" my initial meaning of that is that the hook itself might be re-purposed, perhaps moved into a different module or generally exported so it can be re-used. Actually, though, potentially having an actual histogram widget here as well may be better, with some different presentation options so that it can have the gradient as a background with the distribution visible on top. Not sure how much work it'd take to get that working well, but I think it could likely result in less new code being added in this PR and better results.

nawabfurquan:
A future PR should address the issues mentioned in #497 so that this PR doesn't get too dense

src/react/legend/color_legend/colorLegendFilter.ts
```diff
@@ -57,6 +77,20 @@ export function getActiveCategoricalColorLegendValue(
     return chart.colorLegendFilter.value;
 }
 
+export function getActiveContinuousColorLegendRange(
+    chart: ColorLegendFilterChart,
```
xinaesthete, line 81: We might be wanting to move towards things that are set on the 'layer' level (or for different elements in a chart-array grid) rather than chart-wide, which might mean some of this gets refactored in the not-too-distant future, but I think that's ok.

PR #510, opened 2026-06-19
feature/Jobs POC with `concat_columns` as the first tool/function. 

MarShaikh:
This PR adds a local job execution proof of concept for `mdvtools.jobs`, with end-to-end support for job workspaces, persistence, and worker subprocess execution.

### Key changes:

- Introduces a job workspace abstraction and job store for recording and resuming jobs.
- Adds a pluggable execution backend with `LocalSubProcessExecutor`.
- Adds a worker subprocess entrypoint so any worker can be invoked via `python -m mdvtools.jobs.run_worker "module:fn" <workspace>`.
- Adds worker initialisation and manager logic to track worker `STATUS` and mark job completion.
- Adds support for writing job results back to the datasource.
- Adds provenance build support and provenance tests.
- Adds demo scripting to show intermediary files as they are deleted after job completion.
- Adds coverage for ingest/idempotency, manager tests, job store tests, workspace tests, and concat worker functionality.


MarShaikh:
Hi @xinaesthete please feel free to ignore the ADR files. I think, it would also be a good idea if I add some flow charts to describe how everything connects to each other. I also haven't added provenance to the `datasources.json` file directly, thought that warranted some discussion, though it should be trivial to add it once we finalise what to do, there is a script that records it. Feel free to question/suggest improvements. Thanks.

python/mdvtools/jobs/executor.py
```diff
@@ -0,0 +1,49 @@
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Protocol
+import subprocess
+import sys
+
+
+@dataclass
+class Handle:
+    kind: str  # 'local' | 'slurm' | 'k8s'
```
xinaesthete, line 22: We can probably use `Literal['local', 'slurm', 'k8s']` here.

python/mdvtools/jobs/executor.py
```diff
@@ -0,0 +1,49 @@
...
+    ref: str  # pid | slurm-id | pod-name
+
+
+class Executor(Protocol):
+    def submit(self, entrypoint: str, workspace: Path) -> Handle: ...
```
xinaesthete, line 27: We probably need some way to pass an abort signal into this?

I see that cancellation is documented elsewhere as not being complete at this stage, that seems reasonable, maybe we could try to make this protocol have some approximate shape for it, but maybe it's better to defer. May just be worth a comment here.

python/mdvtools/jobs/ingest.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def ingest_column_output(project, params: dict, ws) -> dict | None:
+    """Output shape column (ds, cols): write the worker's result column on to the datasouce
+    set_column is idempotent - it deletes any existing column of the same name before adding, and updates-or-appends
+    the metadata entry. So re-ingesting the same output_name REPLACES rather than duplicates (ADR0006)."""
+    out_name = params["output_name"]
+    with h5py.File(ws.output / "result.h5", "r") as f:
+        values = [x.decode() for x in f[out_name][:]]  # dataset comes back as bytes
+    project.set_column(params["datasource"], out_name, values)
```
xinaesthete: Happy to see that the use of `h5` is for things which the jobs system owns, and that if the project implementation changes, as long as the public `set_column` interface is stable this should still work.

xinaesthete (commented):
Seems to look reasonably sane on the face of it, haven't tried actually running it.

Also I noticed there was some pyright failing in CI noise, if you could check that. You can `make type` from the python folder to check locally.

MarShaikh:
> Seems to look reasonably sane on the face of it, haven't tried actually running it.
> 
> Also I noticed there was some pyright failing in CI noise, if you could check that. You can `make type` from the python folder to check locally.

Yes, thanks, I'll do that.

python/mdvtools/jobs/executor.py
MarShaikh: Makes sense, I thought maybe it was a bit premature to add fixed typing to this, especially because I don't think we would work with k8s?

xinaesthete:
Thanks for cleaning up that type error, what would you like me to review @MarShaikh - is it worth me running this locally and checking that it behaves? In which case, can you provide some simple instructions?

MarShaikh:
> Thanks for cleaning up that type error, what would you like me to review @MarShaikh - is it worth me running this locally and checking that it behaves? In which case, can you provide some simple instructions?

I actually don't think it's worth running locally, apologies, I forgot this was a draft PR. Maybe it's worth discussing how/where we could integrate this on the frontend? I'm not sure if this is something we want to be including in this draft PR?

xinaesthete:
From our brief conversation just now, I think that the only thing which may run into issues with our current system is the way that it puts a `jobs` folder as a sibling of the project; the catalog is liable to try to scan this as a project and cause some issues. I think we can have `jobs` folder inside the project instead?

For subsequent follow-up is how we structure our types for things like columns so that we don't have several parallel sources-of-truth for that. This also relates to my comment to @nawabfurquan here https://github.com/Taylor-CCB-Group/MDV/pull/512#discussion_r3453745966 (but I think it would be a mistake to try to tackle that here)

I think it's definitely good that we don't bloat `datasources.json` too much, especially with things that are experimental, but we might considering adding some light provisional `provenance` id on column that can be used to reference the `id.json` in jobs here.

p.s. I think just for reference, a very brief high-level note describing `python -m mdvtools.jobs.run_worker "module:fn" <workspace>` would be useful.

PR #521, opened 2026-07-06
Feature/specify datasource

bioinfbloke:
This feature allows the user to not have to always rely on ChatMDV choosing the optimum datasource(s) as this can get complicated on projects with lots of tables. 

The user will see 'Auto' as default when they do a query but you can also toggle this off and choose individual/multiple tables for ChatMDV to focus on when querying the datasources.

## Summary by CodeRabbit

* **New Features**
  * Added multi-datasource selection to chat with **auto** and **manual** modes (including multi-selection) and UI display of chosen/resolved datasources.
  * Extended the chat API/CLI to support datasource-scoped analysis and added resolved datasource names to chat responses.
  * Upgraded guidance generation with richer, chart-aware explanations and more targeted next-step tasks.
* **Bug Fixes**
  * Updated local AI connection configuration to use the standard host URL for Ollama.
* **Tests**
  * Expanded unit/integration coverage for datasource resolution, UI payload shaping, guidance sections, and saved-view chart parsing.

### DIFF
--- a/python/mdvtools/llm/dataset_scale.py
+++ b/python/mdvtools/llm/dataset_scale.py
@@ -139,73 +139,101 @@ def agent_expression_probe_columns(
     return picked or ([name_column] if name_column else [])
 
 
+_AGENT_PROBE_CAP = 6
+
+
+def _load_probe_df_for_datasource(
+    project: Any,
+    datasource_name: str,
+    scale: ProjectScale,
+    roles: Any,
+) -> Any | None:
+    """Load one probe dataframe for the pandas agent (column subset on large projects)."""
+    expr_names = {e.datasource_name for e in (roles.expressions or [])}
+    primary_expr = roles.preferred_expression()
+    is_expression = datasource_name in expr_names
+
+    if scale.is_large:
+        if is_expression and primary_expr and datasource_name == primary_expr.datasource_name:
+            cols = agent_expression_probe_columns(
+                project, datasource_name, primary_expr.name_column
+            )
+        else:
+            cols = agent_probe_columns_from_metadata(project, datasource_name)
+        if not cols:
+            return None
+        return project.get_datasource_as_dataframe(datasource_name, columns=cols)
+
+    return project.get_datasource_as_dataframe(datasource_name)
+
+
 def load_agent_dataframes(
     project: Any,
     roles: Any,
     scale: ProjectScale,
     *,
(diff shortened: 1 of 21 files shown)


python/mdvtools/tests/test_datasource_roles.py
```diff
@@ -447,3 +452,123 @@ def test_rag_prompt_includes_all_tables_and_resolved_datasource():
...
+
+    project = CellsRnaProject()
+    resolved = resolve_datasources_from_question_auto(
+        project,
+        "What are the top marker genes by cluster?",
+    )
+    assert resolved.source == "auto"
+    assert resolved.selected == ["cells", "rna"]
+    assert resolved.primary == "cells"
+
```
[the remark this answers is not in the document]
bioinfbloke: There is also already a regression-style test covering the exact scenario described: catalog order differs from the expected primary, and the result keeps the primary first on the exact datasource-name match path:

nawabfurquan (approved):
I have tried it out and looked at the code, I don't see any major issues
Only minor issue I can see is that the size of conversations is reduced due to those options being added, but we could address that in a future PR, not a blocker

python/mdvtools/llm/dataset_scale.py
```diff
@@ -139,73 +139,110 @@ def agent_expression_probe_columns(
...
 
+_AGENT_PROBE_CAP = 6
+
+
+def _load_probe_df_for_datasource(
+    project: Any,
+    datasource_name: str,
+    scale: ProjectScale,
+    roles: Any,
+) -> Any | None:
```
xinaesthete, line 150: Seems like we know the return type better than `Any | None` here. Could be inferred which would probably be correct type rather than widening.

python/mdvtools/llm/verification.py
```diff
@@ -84,6 +85,70 @@ def _to_param_list(p: Any) -> list[Any]:
     return [p]
 
 
+# Stable role labels for saved-view chart params (by persisted `chart['type']`).
+CHART_PARAM_ROLE_MAP: dict[str, list[str]] = {
```
xinaesthete, line 89: We really need to prioritise having a proper schema-driven understanding of chart configs.

nawabfurquan:
These should hopefully be addressed in the future PRs, merging this now

PR #535, opened 2026-08-12
Generalize spatial table association, and give points layers a real panel

xinaesthete:
Generalises spatial table association across element types, and gives points layers a real panel — feature catalog, per-feature selection and colour, on-demand loading — on top of `@spatialdata` 0.8.0.

Still to be done is improving general look-and-feel / UX of spatial layers controls, probably trying to move into "settings dialog".

## Table association

`table_association.ts` used to speak only shapes. It now covers shapes, labels and points through one path, with `TableAssociationControls` shared by the layer panels instead of each panel growing its own copy. Fill-colour-by-column moved out to `fill_color_scheme.ts` and follows the active link, so a column's colours come from the same place whether MDV or the viewer is asking. `field_spec_projection.ts` keeps MDV's field specs out of the configs handed upstream.

## Points layers

`points_layer_registry.ts` publishes the live `PointsDataEngine` on the chart, and the layer dialog — a portal with its own React tree — bridges it back through `PointsFeatureStateProvider`. Without that the panel would spin up a second engine and answer different questions from the one that is drawing.

`PointsFeatureFilterPanel` lists the feature catalog with per-feature selection, colour overrides, hover highlighting and point counts. Rows say what state they are in: not loaded, partly loaded, or fully rendered. `ensurePointsWorker.ts` starts the core points worker so selecting a feature that is outside the resident window actually fetches it, rather than silently filtering the capped batch.

Four upstream bugs were found and fixed while building this, and this branch takes the release that carries them:

- SpatialData.js#148 — the published worker entries were CJS inside an ESM package, so no feature-index scan could run outside the repo. This app shipped in exactly that state.
- SpatialData.js#149 — a failed scan went quiet and the partial view read as the complete answer.
- SpatialData.js#151 — resident per-feature counts were computed in the wrong code space, so 206 of 541 features claimed more resident points than the dataset contains.
- SpatialData.js#147 — `colorByFeature: false` never reaches the deck layer. Still open; the panel deliberately offers no toggle rather than an inert one.

## Performance

Two regressions found by profiling, both with before/after numbers in their commit messages:

- A cosmetic layer edit re-decoded the shapes geometry. `renderStackPropsGeneration` now separates prop edits from structural ones.
- An opacity drag re-rendered all 541 feature rows — 350–450ms per step, against ~10ms for the deck update the drag was for. `touchRenderStackEntry` subscribes the dialog to every prop key, which is its job, and `spatialEntryAsLayerConfig` then spreads a fresh object, so nothing downstream could tell a cosmetic edit from a real one. The panel now takes a narrowed `PointsFeatureFilterConfig` of the three fields it reads. Narrowing the type rather than leaving a comment means a future edit that reads `pointSize` fails to compile instead of quietly restoring the coupling. Steps now measure 11–14ms.

## Panel layout

The points panel led with point size, colour and memory cap before the feature list anyone opened it for. The colour control was two controls and neither earned its place: the alpha slider wrote `color[3]`, which multiplies against the render stack's own `opacity` prop driven by the layer row slider — two sliders for one visual result. The swatch survives but only matters when there is no feature catalog, so it sits under Advanced when there is one. The memory cap moves into that collapsed section, with its in-memory readout kept in the accordion summary: collapsing should hide the control, not the fact that the view is capped.

## Verification

`vitest run` — 431 passed across 59 files, including new suites for table association, fill-colour schemes and field-spec projection. Behaviour checked by hand against the Xenium pancreas project (8.07M transcripts, 541 features, 4M resident cap): feature selection, colour overrides, hover highlight, on-demand loading of a feature outside the cap, and the Advanced readout.

Two things reviewers should know rather than discover:

- `docs/spatialdata-vis-integration.md` sets the floor at `@spatialdata` >= 0.8.0. Below that the points worker cannot start and on-demand feature loading silently does nothing.
- `scripts/link-spatialdata.mjs` is for local development against a SpatialData.js checkout. The branch itself takes the published packages.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added table associations for spatial shapes and labels, enabling data-driven coloring, filtering, tooltips, and feature visibility.
  - Added an enhanced points-layer panel with feature search, selection controls, color overrides, memory status, and loading diagnostics.
  - Added flexible tooltip and fill-color column selection with support for clearing optional values.
  - Added clearer spatial rendering error messages and improved dialog stacking behavior.

- **Documentation**
  - Expanded guidance for spatial-table integration, performance diagnostics, linked data, and local development workflows.

- **Tests**
  - Added coverage for table associations, color schemes, field projection, and render-stack updates.


xinaesthete:
Worked through the review in b9951687. Taken and declined below, with reasons for the declines.

## Taken

Two were real bugs:

- **Tooltip not cleared on an unpicked hover.** Confirmed — `handleHover` returned before `onFeatureHover` could run, and the only other clear is `onMouseLeave` on the container, so the stale tooltip followed the pointer around the canvas. Added an explicit `onFeatureHoverEnd`.
- **Cleared last colour override written back as `{}`.** Confirmed and reproduced on the real interleaving: recolour and reset inside one JS turn, before the coalescing rAF fires, leaves the pending map empty and `if (pending)` is true for `{}`. Now normalised to `undefined`; verified in the app that `featureColorOverrides` comes out undefined.

Also taken: `ErrorBoundary` `resetKeys`; `useId` for the DataStore listener key; pruning `featureStateCacheRef` / `lastFillColorsRef` for removed layers; `useElementTableAssociation` accepting `elementType: undefined` instead of the placeholder `"shapes"`; the path-aware checkout containment test; `worker_format` validation; focus handlers on feature rows; the doc version floors; `DEFAULT_POINT_COLOR`; and `slotProps.transition` (verified the details stay unmounted while collapsed).

On **zarrextra**: pinned to `0.4.0`, and the link script restores the same pin — `@spatialdata/core@0.8.0` pins it exactly, so `^0.4.0` would let a future `0.4.1` install a second copy. The rest of that comment does not apply here: this repo has no `pnpm-workspace.yaml`, and its `package.json#pnpm.overrides` block is being honoured by pnpm 10.21.0.

## Declined

**`association.project` cache invalidation on a cosmetic edit** (`table_association.ts:725`, rated Major). This one is measurably wrong. Every key the cache compares — `source`, `visibleRows`, `rowCount`, `baseFeatureState`, `colorForRow`, `alpha` — is either a memo output or a value copied by reference through the layer spread, so none of them move on a cosmetic edit. A capture during an opacity drag gives `association.project` count 10, avgMs 0.00, which is the cache holding. The comment above that block already documents why the copy is per-render while the per-feature pass is not.

**Preserve the fill configuration for an unresolved active link** (`TableAssociationControls.tsx:162`). Deliberate, and the reason is in the comment directly above the line: the spec is stored, and `field_spec_projection` fills the colour column in when the link resolves. Writing a stale `columnName` through in the meantime is the behaviour that projection exists to avoid.

**Preserve pre-existing pnpm overrides in the link script.** Real, but the fix is worse than the problem: link and unlink are separate process invocations, so remembering a prior value means writing bookkeeping state into `package.json`. The six names involved are ones this script owns.

**`loadColumnSet` error handling.** `ChartManager.loadColumnSet(columns, dataSource, callback)` has no error channel — there is nothing to catch short of changing ChartManager, which is out of scope here.

Skipped as pure style: the `as` casts in `perf.ts` / `fill_color_scheme.ts` and the normalisation helper in `render_stack_adapter.ts`.

## Not verified

The tooltip fix is correct by inspection but I could not exercise it end to end — I could not get a feature tooltip to appear at all in this environment, which is its own problem and is now filed separately.

`vitest run` 431 passed / 59 files; `tsc --noEmit` clean; biome clean on the changed files.

xinaesthete:
@coderabbitai review

PR #541, opened 2026-08-26
Adopt @spatialdata/* 0.10.0: parquet worker rename, drop parquet-wasm workaround

xinaesthete:
Adopts [SpatialData.js#166](https://github.com/Taylor-CCB-Group/SpatialData.js/pull/166), released as `@spatialdata/*` 0.9.0, and deletes the workaround #539 added.

Core 0.9.0 renamed the points worker to the **parquet worker** with no aliases (it decodes and scans parquet for shapes as much as for points), and loads its vendored parquet-wasm through the `@spatialdata/core/parquet-wasm` export rather than a `@vite-ignore`d relative path no consumer bundler could resolve.

## The rename

`ensurePointsWorker.ts` → `ensureParquetWorker.ts`, whose body is now a single `ensureWorkers()` call from `@spatialdata/vis` — the seam that hides how differently the two workers are wired (the codec worker needs no configuration, the parquet worker needs a URL our bundler produces). It is idempotent and attempts the worker once per page, so MDV's own `pointsWorkerEnabled` latch goes, and it returns whether the worker is actually running.

`PointsFeatureFilterPanel` gates on `isParquetWorkerEnabled()`. From 0.9.0 that answer is trustworthy rather than optimistic: a worker that errors before answering anything is recognised as never wired up and switched off, so callers take their main-thread fallbacks instead of stalling a request at a time. `loadPointsMatchingFeatureCodes` remains the one call with no fallback, which is why the panel still gates on it.

## Deleting the workaround

- `copySpatialdataParquetWasm()` and its `plugins` entry in `vite.config.mts` — the bundler emits the wasm itself now.
- The `/flask/vendor/` auth whitelist entry it needed. Nothing else served from `vendor/`.
- The stale half of the `optimizeDeps.exclude` comment. The Zod 3-vs-4 reason and the CJS-transitives warning both still stand, so it was edited rather than removed.

`bundleAssetReferences.test.ts` stays — it is the right guard regardless, and it now passes because the build emits the asset rather than because a plugin copied one in.

## Two things found while measuring

**`.wasm` was being emitted twice.** `flaskAssetFileNames` routed it to an unhashed `img/parquet_wasm_bg.wasm` for the main graph, while the worker graph (which does not take that function) emitted a hashed copy under `assets/` — 6.6 MB of identical bytes, twice. Matching the worker default collapses them onto one file: `dist/flask` 92M → 86M.

**Subpath imports were broken under `pnpm link:spatialdata`.** Vite matches a *string* alias only on an exact hit or a `/` boundary, so `@spatialdata/core/parquet-worker?worker&url` — the query is part of the id — missed the subpath entry and fell through to the bare `@spatialdata/core → <pkgRoot>` alias, resolving to a path that does not exist. Alias entries are now regexes derived from each linked package's own `exports` map, anchored to end-or-`?` so the query carries through. This was already broken on `main` for `zarrextra/workers`; it only ever bites while testing an upstream change, which is exactly when you need it.

The test suite has one documented inert reference now: core's Node-only `initSync` branch keeps a `../vendor/…wasm` string in the bundle, guarded by `typeof process !== 'undefined' && process.versions?.node != null && typeof window === 'undefined'`, which no browser or worker satisfies.

## Verification

Both a registry build (`build-flask-dockerjs`) and a build against a linked local checkout produce:

- one hashed `assets/parquet_wasm_bg-B1vu2ek6.wasm` (6.6 MB), not two
- `assets/parquet-worker-<hash>.js` (~300 kB)
- no `vendor/` directory

Dev server resolves `@spatialdata/core/parquet-worker?worker&url` and serves it 200. 432 unit tests pass, `tsgo --noEmit` clean, biome clean on the changed files (the repo's 20 pre-existing lint findings are unchanged).

Not yet exercised: a spatial chart loading shapes/points against a real dataset in the devcontainer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Improved on-demand feature loading with the Parquet worker.
  - Added fallback handling when worker startup fails.
  - Improved progressive loading and feature-list performance.

- **Bug Fixes**
  - Disabled feature scanning actions when the required worker is unavailable.
  - Updated authentication enforcement for vendor routes.

- **Documentation**
  - Updated SpatialData integration guidance for current versions and worker setup.

- **Chores**
  - Updated SpatialData components and improved bundled asset handling.


xinaesthete:
Rebased onto `main` (picking up #542) and bumped to **`@spatialdata/* 0.10.0`**.

**Why 0.10.0 and not 0.9.0.** Nothing in 0.10.0 is breaking — the hard compile floor is still 0.9.0, for the parquet worker rename. It is pinned because it is the release where a points layer stopped freezing the tab.

While reviewing this branch I found that the progressive points preload was driving parquet-wasm's `ParquetFile.stream()` on the main thread from *above* the worker gate, so enabling the parquet worker did nothing for it. That is now fixed upstream in [SpatialData.js#174](https://github.com/Taylor-CCB-Group/SpatialData.js/pull/174): the range-fetch and decode happen in the worker, batches are posted back, and the coloured progressive paint survives while the main thread only copies each batch into its accumulator.

Upstream's measurement, on a 4.83M-row Xenium transcripts element capped at 4M with a 12,448-feature panel:

| | before | after |
|---|---|---|
| worker requests | zero | 62 batches, 4M rows |
| preload completes | never (9.4 min, still going) | 75 s |
| worst single task | 113 s | ~4.2 s |

0.10.0 also virtualizes the *upstream* feature list ([SpatialData.js#172](https://github.com/Taylor-CCB-Group/SpatialData.js/issues/172)) — the same fix #542 made to our own `PointsFeatureFilterPanel`. The two are independent, and on that element neither alone was enough: roughly half the blocking was the decode and half was the DOM.

**The rebase.** #542 rewrote `PointsFeatureFilterPanel.tsx`, which this branch also touches (the `isPointsWorkerEnabled` → `isParquetWorkerEnabled` rename). Both survived cleanly — the panel now has the renamed symbol and the virtualization.

**One thing that did not change.** `bundleAssetReferences.test.ts` keeps its single inert-reference entry. I had said earlier that the upstream `import.meta` fix would let us drop it; that was wrong. The guarded Node-only branch is still compiled into core's ESM build, so the `../vendor/…wasm` literal is still in the bundle — only the cjs side changed, and we bundle the ESM.

`zarrextra` stays 0.5.0; core@0.10.0 still depends on exactly that.

Verified: production build emits one hashed 6.6 MB `parquet_wasm_bg.wasm`, a `parquet-worker` chunk, no `vendor/`; 432 tests pass, `tsgo` clean, biome unchanged at main's 20 pre-existing findings.

